PR #4, opened 2020-04-13
Fixed build with OpenCV 4.3

RazZziel:
Build fails on Archlinux because we don't have OpenCV 2 anymore, only OpenCV 4


floe:
The includes are backwards-compatible, but the Makefile change would break the build with OpenCV 3. Any ideas, maybe something with `pkg-config --exists`?

RazZziel:
I've modified the change, what do you think?

A quick and dirty fallback as in `pkg-config --libs --cflags opencv || pkg-config --libs --cflags opencv4` would probably have been enough, but I felt it was cleaner and more readable to separate everything properly

floe:
Looking good, thanks!

PR #5, opened 2020-04-13
fix for fedora 31

mkogan1:
probably necessary because OpenCV stock package version is opencv-3.4.8
and compiler performance optimization flags


floe:
Thanks, merged manually (the second change with `CV_CAP_MODE_YUYV` doesn't work for me, so I left the raw FOURCC code in).

PR #12, opened 2020-04-28
Parameters for most things

phlash:
command line parameters for: debug, cameras (in/out), size, background, threads, avoids edit/recompile cycles for faster experimentation.


floe:
Excellent, thanks, merged!

PR #13, opened 2020-05-03
Better capture device support

phlash:
Improving capture from IP cameras / phone apps, supporting multiple capture formats:

 * enables CV_CAP_PROP_CONVERT_RGB (actually BGR24, thanks OpenCV!), allowing many devices to work without us knowing exactly what they spit out and handling conversion.
 * background images now usable without colour space conversion (as default is BGR24 too).
 * resizes capture -> output, in case capture devices are weird (eg: mjpeg from my phone is two pixels off at 638x478??)
 * ~~captures in a separate thread to avoid video lag if TF isn't quick enough (it usually isn't!)~~

I've had a fun Saturday ;)

### DIFF
--- a/deepseg.cc
+++ b/deepseg.cc
@@ -27,28 +27,6 @@ limitations under the License.
 
 #include "loopback.h"
 
-// OpenCV helper functions
-cv::Mat convert_rgb_to_yuyv( cv::Mat input ) {
-	cv::Mat tmp;
-	cv::cvtColor(input,tmp,CV_RGB2YUV);
-	std::vector<cv::Mat> yuv;
-	cv::split(tmp,yuv);
-	cv::Mat yuyv(tmp.rows, tmp.cols, CV_8UC2);
-	uint8_t* outdata = (uint8_t*)yuyv.data;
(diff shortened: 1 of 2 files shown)


floe:
Wow, that's a big one :-) Do you think you can split it, one PR for the threading stuff and one for the color conversion?

phlash:
Yep, no problem, I'll take out the threading stuff for this PR..

cob16:
:+1: This pr has got `Logitech HD Pro Webcam C920` working for me :smile:

floe:
Sorry for the long silence, I haven't forgotten about this, I'm just absolutely swamped ATM :-/

BenBE (commented):
LGTM. Needs some rebasing for current master and to work together with my PR #36. Would make the mirroring feature from my PR even much saner (avoiding two colorspace conversions that YUYV requires me for horizontal mirroring …

floe:
The main reason why I've been reluctant to merge this one is because it adds a noticeable amount of overhead, at least on my machine, due to the multiple colorspace conversions. OTOH, when using the Google Meet network, there's a bit more headroom as it is way more efficient...

phlash:
Superseded by #48

PR #15, opened 2020-05-04
Threaded capture to keep up with camera and reduce latency

phlash:
This PR adds a capture thread to ensure we keep up with input frame rate, otherwise some OpenCV capture back ends (notably RTSP & MJPEG, both via libavformat/ffmpeg) buffer up and produce increasing latency (I've seen 20+ seconds).


floe:
Would it make sense to use double-buffering, i.e. put two Mat objects into capinfo_t and then avoid the copy in line 222?

phlash:
Possibly, there are a few constraints to meet:
 * capture needs to constantly retrieve images (to a Mat) in order to keep up with the stream (https://stackoverflow.com/questions/36913733/opencv-grab-does-nothing-on-webcam)
 * `Mat raw` needs to remain constant until the main thread completes a processing cycle, since it is used at start and end.

with this in mind, I think it makes sense to have two Mat objects, switched by the main thread under lock: the capture thread always writes to one, the main thread processes the other after a switch. I'll have a go :)

PR #38, opened 2021-01-30
Outsource mask calculation to new thread to decouple from clipping

peckto:
This PR outsources the mask calculation via tensorflow from the actual background clipping process.
In this way we can decouple the (slow) ANN task from the (fast) clipping task.
This results in a real time video rendering, even if the ANN task is slow(er).
As a side effect, the mask might be slower updated then the people moving the the video.
After all, this makes a smoother user experience, even with slower, but more stable models like `deeplabv3_257_mv_gpu`.

### DIFF
--- a/deepseg.cc
+++ b/deepseg.cc
@@ -53,11 +53,11 @@ cv::Mat convert_rgb_to_yuyv( cv::Mat input ) {
 // Tensorflow Lite helper functions
 using namespace tflite;
 
-#define TFLITE_MINIMAL_CHECK(x)                              \
-  if (!(x)) {                                                \
+#define TFLITE_MINIMAL_CHECK(x)                               \
+	if (!(x)) {                                              \
 	fprintf(stderr, "Error at %s:%d\n", __FILE__, __LINE__); \
 	exit(1);                                                 \
-  }
+	}
 
 std::unique_ptr<Interpreter> interpreter;
 
@@ -69,7 +69,7 @@ cv::Mat getTensorMat(int tnum, int debug) {
 	TfLiteIntArray* dims = interpreter->tensor(tnum)->dims;
 	if (debug) for (int i = 0; i < dims->size; i++) printf("tensor #%d: %d\n",tnum,dims->data[i]);
 	TFLITE_MINIMAL_CHECK(dims->data[0] == 1);
-	
+
 	int h = dims->data[1];
 	int w = dims->data[2];
 	int c = dims->data[3];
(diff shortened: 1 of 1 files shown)


floe:
Thanks, good suggestion - but that's a really big PR to review, do you think we can split it (e.g. first refactor the main chunk of code into a separate method, and then introduce the second thread)?

floe:
(Now that I think about it, would it maybe make sense to not introduce a third thread, but rather have the mask calculation just run on the capture thread, and the NN calculation on the main thread?)

deepseg.cc
```diff
@@ -403,21 +450,27 @@ int main(int argc, char* argv[]) {
...
-		fflush(stdout);
-		if (debug < 2) continue;
-
 		cv::Mat test;
 		cv::cvtColor(raw,test,CV_YUV2BGR_YUYV);
 		cv::imshow("output.png",test);
 		if (cv::waitKey(1) == 'q') break;
+
+		// throttle main trhead to save performance for tensorflow threads
+		usleep(100000);
```
BenBE, line 459: Maybe use condition variables or some event monitoring library like libev to handle the multiplexing …

deepseg.cc
```diff
@@ -53,11 +53,11 @@ cv::Mat convert_rgb_to_yuyv( cv::Mat input ) {
 // Tensorflow Lite helper functions
 using namespace tflite;
 
-#define TFLITE_MINIMAL_CHECK(x)                              \
-  if (!(x)) {                                                \
+#define TFLITE_MINIMAL_CHECK(x)                               \
+	if (!(x)) {                                              \
 	fprintf(stderr, "Error at %s:%d\n", __FILE__, __LINE__); \
 	exit(1);                                                 \
```
BenBE, line 59: Shouldn't these gain one level of indentation?

deepseg.cc
```diff
@@ -53,11 +53,11 @@ cv::Mat convert_rgb_to_yuyv( cv::Mat input ) {
 // Tensorflow Lite helper functions
 using namespace tflite;
 
-#define TFLITE_MINIMAL_CHECK(x)                              \
-  if (!(x)) {                                                \
+#define TFLITE_MINIMAL_CHECK(x)                               \
```
BenBE, line 56: Maybe be consistent with the alignment at the right?

peckto:
Thank you for your feedback.
Sure, I will provide a second PR for splitting up the NN and rendering task is separate functions.

About your idea with the capture thread, would be possible. In this way we could get rid of the third thread.
I'm just not so sure, if it's a cleaner separation with the extra thread. 
Maybe we will see after refactoring the code.

peckto:
And sorry for the code style issues, I will check my IDE settings.

floe:
Continuing in #41 for now.

Issue #47, opened 2021-02-19
(-215:Assertion failed) sz.width % 2 == 0 && sz.height % 3 == 0 in function 'CvtHelper'


Rapha149:
This is the output of `./deepseg -d -d -c /dev/video0 -v /dev/video1`:

```
deepseg v0.2.0
(c) 2021 by floe@butterbrot.org
https://github.com/floe/deepbacksub
debug:  2
ccam:   /dev/video0
vcam:   /dev/video1
width:  640
height: 480
flip_h: no
flip_v: no
threads:2
back:   (none)
model:  models/segm_full_v679.tflite

vid_format->type                = 2
vid_format->fmt.pix.width       = 640
vid_format->fmt.pix.height      = 480
vid_format->fmt.pix.pixelformat = 1448695129
vid_format->fmt.pix.sizeimage   = 614400
vid_format->fmt.pix.field       = 1
vid_format->fmt.pix.bytesperline= 1280
vid_format->fmt.pix.colorspace  = 8

tensor #0: 1
tensor #0: 144
tensor #0: 256
tensor #0: 3
tensor #244: 1
tensor #244: 144
tensor #244: 256
tensor #244: 2
Invalid MIT-MAGIC-COOKIE-1 keyQSettings::value: Empty key passed
QSettings::value: Empty key passed
terminate called after throwing an instance of 'cv::Exception'
  what():  OpenCV(3.4.9) /home/abuild/rpmbuild/BUILD/opencv-3.4.9/modules/imgproc/src/color.simd_helpers.hpp:104: error: (-215:Assertion failed) sz.width % 2 == 0 && sz.height % 3 == 0 in function 'CvtHelper'
```

Thanks in advance!

phlash:
Not sure if the `Invalid MIT-MAGIC-COOKIE-1 key` is the problem or the assertion.. if you are trying to run this remotely, you may want to remove the `-d -d` to avoid the application trying to display an image (but not finding a valid .Xauthority file). With the display issue out of the way we have a chance :)

Rapha149:
I don't know what `Invalid MIT-MAGIC-COOKIE-1 key` exactly is, but I have a temporarily fix for it. If I do the fix, the output is the same, just without the `Invalid MIT-MAGIC-COOKIE-1 key`. 
That's the output of the gdb commands:
```
#0  0x00007ffff2ad9520 in raise () at /lib64/libc.so.6
[...]

Issue #46, opened 2021-02-19
'cv::Exception': Invalid number of channels in input image


Rapha149:
Hello,

if I run `./deepseg -d -d -c /dev/video0 -v /dev/video1` this output appears:

```
deepseg v0.2.0
(c) 2021 by floe@butterbrot.org
https://github.com/floe/deepbacksub
debug:  2
ccam:   /dev/video0
vcam:   /dev/video1
width:  640
height: 480
flip_h: no
flip_v: no
threads:2
back:   (none)
model:  models/segm_full_v679.tflite

vid_format->type                = 2
vid_format->fmt.pix.width       = 640
vid_format->fmt.pix.height      = 480
vid_format->fmt.pix.pixelformat = 1448695129
vid_format->fmt.pix.sizeimage   = 614400
vid_format->fmt.pix.field       = 1
vid_format->fmt.pix.bytesperline= 1280
vid_format->fmt.pix.colorspace  = 8

tensor #0: 1
tensor #0: 144
tensor #0: 256
tensor #0: 3
tensor #244: 1
tensor #244: 144
tensor #244: 256
tensor #244: 2
terminate called after throwing an instance of 'cv::Exception'
  what():  OpenCV(3.4.9) /home/abuild/rpmbuild/BUILD/opencv-3.4.9/modules/imgproc/src/color.simd_helpers.hpp:88: error: (-2:Unspecified error) in function 'cv::impl::{anonymous}::CvtHelper<VScn, VDcn, VDepth, sizePolicy>::CvtHelper(cv::InputArray, cv::OutputArray, int) [with VScn = cv::impl::{anonymous}::Set<2>; VDcn = cv::impl::{anonymous}::Set<3, 4>; VDepth = cv::impl::{anonymous}::Set<0>; cv::impl::{anonymous}::SizePolicy sizePolicy = (cv::impl::<unnamed>::SizePolicy)2; cv::InputArray = const cv::_InputArray&; cv::OutputArray = const cv::_OutputArray&]'
> Invalid number of channels in input image:
>     'VScn::contains(scn)'
> where
>     'scn' is 1
```

Thanks in advance!

phlash:
Possible duplicate of #3 ?

@Rapha149 - can you check your camera output formats as per the earlier ticket, this might help identify if it's a variant of the same issue. Ta :)

Rapha149:
I think you meant this command: `v4l2-ctl --list-formats-ext`, am I right?
The output of this command is
```
[...]

Issue #34, opened 2021-01-21
Improve temporal stability of masked image


BenBE:
The masked image is often quite unstable in regards to the flow of time. While certain details may appear in one frame they might be gone in the next frame. This is in particular distracting if this affects part of the body or face (not necessarily the borders).

Maybe decrease the likelyhood of a pixel being removed from the mask on the amount of pixels that changed mask membership in their neighborhood: I.e. the more pixels in an area suddently change membership, the less likely it is that the whole area should be removed.

floe:
I have two ideas for solving this (which are non-exclusive):

- use a weighted moving average over the last n frames
- use a outline/hull function to close "holes" inside the silhouette

Will probably try to implement this soon-ish (exam period ATM). Pull requests welcome, of course :-)

BenBE:
The days when I had image processing in university are unfortunately quite a few years back …

For the weighted average: Combine with gamma correction of likelyhood detection by the NN. With the hard segmentation (softmax) only after the averaging …

Both seem reasonable although my gut feeling suggests the first option may be more promising when trying to avoid the flicker of the misdetections.

P.S.: Good luck with your exams.

phlash:
Seems related to #29 with the segm models?

BenBE:
Similar, but not quite. This issue is for WHEN the model segments a proper image that it sometimes (for only a few frames) removes large portions of the segmentation.

#29 is primarily about the model not segmenting correctly in the first place.

With #29 resolved you'd get a proper image, which still might flicker from time to time.
With this image result, you might not get the proper image, but the part it shows is consistent over time.

floe:
WMA was pretty straightforward to implement and does look noticeably better, IMHO.

PR #48, opened 2021-02-19
Format Conversion issues causing OpenCV exceptions

phlash:
Fix for issues #46 and #47, using the OpenCV capture code to perform BGR conversion instead of trying to do it ourselves, appears to support more cameras. This also supersedes #13.

### DIFF
--- a/deepseg.cc
+++ b/deepseg.cc
@@ -245,8 +245,7 @@ int main(int argc, char* argv[]) {
 
 	cap.set(CV_CAP_PROP_FRAME_WIDTH,  width);
 	cap.set(CV_CAP_PROP_FRAME_HEIGHT, height);
-	cap.set(CV_CAP_PROP_FOURCC, *((uint32_t*)"YUYV"));
-	cap.set(CV_CAP_PROP_CONVERT_RGB, false);
+	cap.set(CV_CAP_PROP_CONVERT_RGB, true);
 
 	// Load model
 	std::unique_ptr<tflite::FlatBufferModel> model =
@@ -321,9 +320,9 @@ int main(int argc, char* argv[]) {
 		cv::Mat roi = raw(roidim);
 
 		// resize ROI to input size
-		cv::Mat in_u8_yuv, in_u8_rgb;
-		cv::resize(roi,in_u8_yuv,cv::Size(input.cols,input.rows));
-		cv::cvtColor(in_u8_yuv,in_u8_rgb,CV_YUV2RGB_YUYV);
+		cv::Mat in_u8_bgr, in_u8_rgb;
+		cv::resize(roi,in_u8_bgr,cv::Size(input.cols,input.rows));
+		cv::cvtColor(in_u8_bgr,in_u8_rgb,CV_BGR2RGB);
 		// TODO: can convert directly to float?
 
 		// bilateral filter to reduce noise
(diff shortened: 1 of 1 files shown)


BenBE (commented):
General patch idea LGTM. But can't say anything decisively about the performance impact due to these conversions.

As TF want's images as `RGB_float32` anyway it might be worth thinking about doing that conversion for the captured source image once (and any supplementary images for backgrounds/animations) and do just a second conversion to YUYV right before writing the image back to the `v4l2loopback` device.

Doing so should make several operations perform much better on the GPU (especially masking/blending) but should also allow for real "exponential falloff" for time-smoothened buffers (cf. #34).

As I can hardly give a good assessment on the impact of this patch, I'd suggest handing this of to @floe.

phlash:
As we seem to be unsure of timing/performance impacts, I added nanosecond timing capture through the main loop, and in the grabber thread, and ran both the original and patched code as above on my non-GPU accelerated laptop with the default Google Meet segmentation model. Results in the attached file, and my summary:

 * This BGR input patch increases the grabber threads retrieve() time in openCV from ~0.07ms to ~4.0ms. No impact to FPS as this is in a separate thread, but an increase in CPU usage.
 * The majority of processing time is in TFlite (20-30ms), this _may_ be quicker on a GPU accelerated device (I cannot test).
 * openCV pre-processing is 3.5-4.0ms, I will break this down into smaller steps next.
 * mask calculation is ~0.55ms
 * openCV post-processing is ~0.6ms
 * V4L2 writes are ~0.07ms
 * with the patch I turn into a green blob, so there is something not quite right in the masking/output with a BGR captured frame.

[timings.txt](https://github.com/floe/deepbacksub/files/6015027/timings.txt)

timings points are as follows:
 * Startup - entry point to start of main loop
 * wait - top of loop to mutex lock
 * lock - duration of mutex lock
 * grab - from grabber thread, duration of grab() call
 * retr - from grabber thread, duration of retrieve() call
 * copy - duration of copy to local cv::Mat raw
 * open - all remaining openCV pre-processing to TFlite call
 * tflt - duration of TFlite invoke call
 * mask - calculation of output mask from TFlite results
 * post - all openCV post-processing and mask apply
 * v4l2 - write to V4L2 loopback device

phlash:
Green blob fixed - the background image was YUYV which messed up the BGR raw capture when copied into it.

BenBE:
Can you split the timing commit into its own PR? I think we should keep that part independent of the other changes.

BenBE:
I'd prefer if we can keep the patch series per issue as short as reasonable.

For that reason I dropped those two intermediate commits and rebased it to `main`. 

Regarding the timing you provided: Overall it looks that applying that patch series reduces the stddev between frames (i.e. reduces processing jitter), but overall slightly increases the frame processing time (causing a drop by 0.8fps). While this is still acceptable with your base FTP of >30fps it's somewhat noticeable on e.g. my hardware with usually only about 15fps.

Overall we should start to do some performance analysis of where those extreme slow processing steps come from and if we can optimize them.

I'd opt for including this PR due to the cleanup of the flip code it allows; but probably only on the experimental branch.

NB: We should put #49 on main first, fast-forward experimental to main and merge this PR to experimental then. That way we have a solid base for the next cleanup steps.

floe:
Thanks for your efforts, guys! In general, I agree that using RGB throughout is a nicer solution with less compatibility issues, and a difference of 0.8 FPS is probably not worth the hassle of dealing with YUV all the time.

floe (approved):

PR #49, opened 2021-02-20
added timing details

phlash:
Timing details for main processing loop and frame grabber thread, to help with understanding impacts of logic changes on CPU loading and FPS.

### DIFF
--- a/deepseg.cc
+++ b/deepseg.cc
@@ -89,8 +89,17 @@ typedef struct {
 	cv::Mat *grab;
 	cv::Mat *raw;
 	int64 cnt;
+	long grabns;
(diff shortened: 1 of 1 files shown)


BenBE (changes requested):
Please put the timings into their own struct, to avoid littering the local variable namespace with too many timing-related variables. Similar can be done for the grabber thread, so that the timing information from there can easily be copied over to the main timing context.

BenBE (commented):
LGTM.

As we are with C++, it'd probably make sense to use `std::chrono` too.

BenBE (approved):

PR #50, opened 2021-02-21
FEATURE: Add CMake based build

oleid:
Thanks for this cool project!

Tensorflow-lite can now be used as a cmake subproject, which makes the build more robust and IMHO simplifies it.

This is a cmake project file using the aforementioned subproject feature. Hope you find it useful!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,31 @@
+cmake_minimum_required(VERSION 3.16)
+project(deepbacksub CXX)
+
+find_package(OpenCV REQUIRED COMPONENTS core imgproc imgcodecs video videoio highgui)
+
+# This assumes that tensorflow got checked out as submodule.
+# True, if `git clone` had the additional option `--recursive`.
+add_subdirectory(tensorflow/tensorflow/lite 
+  "${CMAKE_CURRENT_BINARY_DIR}/tensorflow-lite" EXCLUDE_FROM_ALL)
+
+add_executable(deepseg
+  deepseg.cc
+  loopback.cc 
+  transpose_conv_bias.cc
+)
+
+target_link_libraries(deepseg 
+      tensorflow-lite ${CMAKE_DL_LIBS}
+      opencv_core 
+      opencv_video
+      opencv_videoio
+      opencv_imgproc
+      opencv_imgcodecs
+      opencv_highgui
+)
+
+
+install(TARGETS deepseg)
+install(DIRECTORY models 
+    DESTINATION ${CMAKE_INSTALL_PREFIX}/share/deepbacksub 
+    FILES_MATCHING PATTERN "*.tflite" PATTERN "*.md")
--- a/README.md
+++ b/README.md
@@ -111,13 +111,18 @@ Tested with the following software:
 
 Install dependencies (`sudo apt install libopencv-dev build-essential v4l2loopback-dkms curl`).
 
+Clone this repository with `git clone --recursive https://github.com/floe/deepbacksub.git`.
+To speed up the checkout you can additionally pass `--depth=1` to `git clone`.
+This is okay, if you only want to download and build the code, however, for development it is not recommended.
+
 Run `make` to build everything (should also clone and build Tensorflow Lite).
+Alternatively, you can also use `cmake`. Create a subfolder (e.g. `build`), change to that folder and run: `cmake && make -j4` (please replace "4" with the number of your cores).
 
 If the first part doesn't work:
   - Clone https://github.com/tensorflow/tensorflow/ repo into `tensorflow/` folder
   - Checkout tag v2.4.0
-  - run ./tensorflow/lite/tools/make/download_dependencies.sh
-  - run ./tensorflow/lite/tools/make/build_lib.sh
+  - run ./tensorflow/lite/tools/make/download_dependencies.sh (only for non-cmake build)
+  - run ./tensorflow/lite/tools/make/build_lib.sh (only for non-cmake build)
 
 ## Usage
 


README.md
```diff
@@ -111,13 +111,16 @@ Tested with the following software:
 
 Install dependencies (`sudo apt install libopencv-dev build-essential v4l2loopback-dkms curl`).
 
+Clone this repository with `git clone --recursive --depth=1 https://github.com/floe/deepbacksub.git`.
+
 Run `make` to build everything (should also clone and build Tensorflow Lite).
+Alternatively, you can also use `cmake`. Create a subfolder (e.g. `build`), change to that folder and run: `cmake -GNinja .. && ninja`.
```
BenBE, line 117: I know there are multiple generators for `cmake`, so using `ninja` is not required. Thus I'd refrain from explicitly suggesting it, because this looks like yet another dependency that a normal user might want to avoid. The default generator of cmake should be just fine, thus why not stick with it?

README.md
```diff
@@ -111,13 +111,16 @@ Tested with the following software:
 
 Install dependencies (`sudo apt install libopencv-dev build-essential v4l2loopback-dkms curl`).
 
+Clone this repository with `git clone --recursive --depth=1 https://github.com/floe/deepbacksub.git`.
```
BenBE, line 114: Using shallow clones comes with its quite own downsides (even though most of them have been addressed in recent `git` releases). Please drop the `--depth=1` in the example snippet. It can be mentioned in a second sentence, but shouldn't be the default that we document.

CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
+cmake_minimum_required(VERSION 3.16)
+project(deepbacksub CXX)
+
+find_package(OpenCV REQUIRED COMPONENTS core imgproc video videoio highgui)
+
+# This assumes that tensorflow got checked out as submodule.
+# True, if `git clone` had the additional option `--recursive`.
+add_subdirectory(tensorflow/tensorflow/lite 
+  "${CMAKE_CURRENT_BINARY_DIR}/tensorflow-lite" EXCLUDE_FROM_ALL)
```
BenBE, line 9: Why are you setting the `EXCLUDE_FROM_ALL` property here? Any reasonable makefile generator should build the build instructions in a way such that tensorflow won't be rebuilt unless you change anything in it - in which case it should be rebuilt by definition.

BenBE (changes requested):

CMakeLists.txt
oleid: This is actually from the official tensorflow-lite example; my best guess is: there are dependencies, which are not needed for each build. Or even dependencies, which could inject install rules.
In the end, this will allow users to simply issue `make`  instead of `make deepseg` and be done with it.

BenBE: k. Just wanted to make sure there's some reasoning behind this. :)

README.md
oleid: It is slower for large things like tensorflow, but nevertheless, most people know about it thus they will automatically choose it. Changed README to Make generator.

README.md
oleid: Okay, I've adapted README accordingly.

phlash:
Looks like this won't work on vanilla Debian stable (buster), as the default `cmake` package is too old (3.13.4, TF claims to need 3.16). A 3.16.3 version exists in buster-backports, which it's happy with however.

BenBE:
The patch as is causes several unrelated targets to be rebuild repeatedly:

```
…
Scanning dependencies of target deepseg
[100%] Building CXX object CMakeFiles/deepseg.dir/deepseg.cc.o
[100%] Building CXX object CMakeFiles/deepseg.dir/loopback.cc.o
[100%] Building CXX object CMakeFiles/deepseg.dir/transpose_conv_bias.cc.o
[100%] Linking CXX executable deepseg
[100%] Built target deepseg
user@host:~/deepbacksub/build (add_cmake_build[oleid])$ make && ./deepseg -t 8 -m ../models/deeplabv3_257_mv_gpu.tflite
[  0%] Built target pthreadpool
[  0%] Built target absl_spinlock_wait
[  0%] Built target absl_log_severity
[  0%] Built target absl_raw_logging_internal
[  0%] Built target absl_dynamic_annotations
[  0%] Built target absl_throw_delegate
[  0%] Built target absl_base
[  0%] Built target absl_malloc_internal
[  1%] Built target absl_debugging_internal
[  1%] Built target absl_stacktrace
[  1%] Built target absl_demangle_internal
[  1%] Built target absl_symbolize
[  1%] Built target absl_bad_optional_access
[  1%] Built target absl_civil_time
[  1%] Built target absl_int128
[  1%] Built target absl_strings_internal
[  3%] Built target absl_strings
[  3%] Built target absl_time_zone
[  5%] Built target absl_time
[  5%] Built target absl_graphcycles_internal
[  5%] Built target absl_synchronization
[  5%] Built target absl_flags_program_name
[  5%] Built target absl_flags_config
[  5%] Built target absl_str_format_internal
[  5%] Built target absl_flags_marshalling
[  6%] Built target absl_flags_registry
[  6%] Built target absl_flags_internal
[  6%] Built target absl_flags
[  6%] Built target absl_city
[  6%] Built target absl_bad_variant_access
[  6%] Built target absl_hash
[  6%] Built target absl_cord
[  6%] Built target absl_status
[  6%] Built target farmhash
[  6%] Built target fft2d_fftsg
[  6%] Built target fft2d_fftsg2d
[  8%] Built target flatbuffers
[ 10%] Built target ruy
[ 10%] Built target clog
[ 10%] Built target cpuinfo
[ 90%] Built target XNNPACK
[100%] Built target tensorflow-lite
[100%] Built target deepseg
deepseg v0.2.0
(c) 2021 by floe@butterbrot.org
https://github.com/floe/deepbacksub
debug:  2
ccam:   /dev/video0
vcam:   /dev/video1
width:  640
height: 480
flip_h: no
flip_v: no
threads:8
back:   (none)
model:  ../models/deeplabv3_257_mv_gpu.tflite

vid_format->type                = 2
vid_format->fmt.pix.width       = 640
vid_format->fmt.pix.height      = 480
vid_format->fmt.pix.pixelformat = 1448695129
vid_format->fmt.pix.sizeimage   = 614400
vid_format->fmt.pix.field       = 1
vid_format->fmt.pix.bytesperline= 1280
vid_format->fmt.pix.colorspace  = 8

tensor #183: 1
tensor #183: 257
tensor #183: 257
tensor #183: 3
tensor #168: 1
tensor #168: 257
tensor #168: 257
tensor #168: 21
user@host:~/deepbacksub/build (add_cmake_build[oleid])$ make && ./deepseg -t 8 -m ../models/deeplabv3_257_mv_gpu.tflite
[  0%] Built target pthreadpool
[  0%] Built target absl_spinlock_wait
[  0%] Built target absl_log_severity
[  0%] Built target absl_raw_logging_internal
[  0%] Built target absl_dynamic_annotations
[  0%] Built target absl_throw_delegate
[  0%] Built target absl_base
[  0%] Built target absl_malloc_internal
[  1%] Built target absl_debugging_internal
[  1%] Built target absl_stacktrace
[  1%] Built target absl_demangle_internal
[  1%] Built target absl_symbolize
[  1%] Built target absl_bad_optional_access
[  1%] Built target absl_civil_time
[  1%] Built target absl_int128
[  1%] Built target absl_strings_internal
[  3%] Built target absl_strings
[  3%] Built target absl_time_zone
[  5%] Built target absl_time
[  5%] Built target absl_graphcycles_internal
[  5%] Built target absl_synchronization
[  5%] Built target absl_flags_program_name
[  5%] Built target absl_flags_config
[  5%] Built target absl_str_format_internal
[  5%] Built target absl_flags_marshalling
[  6%] Built target absl_flags_registry
[  6%] Built target absl_flags_internal
[  6%] Built target absl_flags
[  6%] Built target absl_city
[  6%] Built target absl_bad_variant_access
[  6%] Built target absl_hash
[  6%] Built target absl_cord
[  6%] Built target absl_status
[  6%] Built target farmhash
[  6%] Built target fft2d_fftsg
[  6%] Built target fft2d_fftsg2d
[  8%] Built target flatbuffers
[ 10%] Built target ruy
[ 10%] Built target clog
[ 10%] Built target cpuinfo
[ 90%] Built target XNNPACK
[100%] Built target tensorflow-lite
[100%] Built target deepseg
deepseg v0.2.0
(c) 2021 by floe@butterbrot.org
https://github.com/floe/deepbacksub
debug:  2
ccam:   /dev/video0
vcam:   /dev/video1
width:  640
height: 480
flip_h: no
flip_v: no
threads:8
back:   (none)
model:  ../models/deeplabv3_257_mv_gpu.tflite

vid_format->type                = 2
vid_format->fmt.pix.width       = 640
vid_format->fmt.pix.height      = 480
vid_format->fmt.pix.pixelformat = 1448695129
vid_format->fmt.pix.sizeimage   = 614400
vid_format->fmt.pix.field       = 1
vid_format->fmt.pix.bytesperline= 1280
vid_format->fmt.pix.colorspace  = 8

tensor #183: 1
tensor #183: 257
tensor #183: 257
tensor #183: 3
tensor #168: 1
tensor #168: 257
tensor #168: 257
tensor #168: 21
```

While I can live with this it's probably not as intended. Also there should be a documented way to configure the tf-lite build (as is possible with the previous GNU make variant, which only relies on the built lib+headers to be present.

oleid:
> The patch as is causes several unrelated targets to be rebuild repeatedly:
> 
> While I can live with this it's probably not as intended. 

Does it take a long time? I only tried this with `ninja` -- it might be faster than `make`.

As for being intended: using `tf-lite` as a subproject is officially documented here: https://www.tensorflow.org/lite/guide/build_cmake (scroll down to the bottom of the page). If this is not as intended, it would be a bug in tensorflow's cmake code.


> Also there should be a documented way to configure the tf-lite build (as is possible with the previous GNU make variant, which only relies on the built lib+headers to be present.

Do you mean "configure" as in externally built,  - build once, re-use artifact, without touching it again?
Or "configure" as in apply certain settings?

The former would be possibly by using CMake's external project support. Which is quite ugly.
The latter would be possibly by using CMake's cached settings. The available options are documented at https://www.tensorflow.org/lite/guide/build_cmake and can either be changed via `ccmake` or running cmake with the corresponding options. I tried to enable `TFLITE_ENABLE_GPU`, but it didn't work for me.

oleid:
> Looks like this won't work on vanilla Debian stable (buster), as the default `cmake` package is too old (3.13.4, TF claims to need 3.16). A 3.16.3 version exists in buster-backports, which it's happy with however.

Yeah, that's a requirement of Tensorflow. I *think* they need such a modern version due to CMake's source code download features.

> confirmed - adding `opencv_imgcodecs` to the dependency list completes the build.

Uh, strange. I didn't need that one on ArchLinux. Probably different things in different libraries for different opencv versions.
I added imgcodecs to the `CMakeLists.txt`

oleid:
> The patch as is causes several unrelated targets to be rebuild repeatedly:

Do you also see this behavior when calling `make deepseg` instead of only calling `make`? I just realized that I don't see that behavior when opening the CMake project in my IDE and just push the run button.

BenBE:
> > The patch as is causes several unrelated targets to be rebuild repeatedly:
> 
> Do you also see this behavior when calling `make deepseg` instead of only calling `make`?

Yes. Same behaviour.

Doesn't take long, but isn't as it should be either.

oleid:
> > > The patch as is causes several unrelated targets to be rebuild repeatedly:
> > 
> > 
> > Do you also see this behavior when calling `make deepseg` instead of only calling `make`?
> 
> Yes. Same behaviour.
> 
> Doesn't take long, but isn't as it should be either.

It might be a limitation of CMake's makefile generator, as I don't see this behaviour when running `ninja deepseg`

```
$ touch ../deepseg.cc
$ ninja deepseg
[1/2] Building CXX object CMakeFiles/deepseg.dir/deepseg.cc.o
[2/2] Linking CXX executable deepseg
```

phlash:
I dug into this a bit today - it's a result of the `add_subdirectory` commands for each dependency listed by Tensorflow Lite, which in turn is a subdirectory of `backscrub` itself. Sensibly, make descends into each sub directory and checks everything is up to date, but for whatever reason, CMake also appends a progress printing line after each descent, as found in the `CMakeFiles/Makefile2`:
```make
<sub-directory>:
    $(MAKE) -f CMakeFiles/<sub-directory>.dir/build.make CMakeFiles/<sub-directory.dir/build
    @$(CMAKE_COMMAND) -E cmake_echo_color ... "Built target <sub-directory>"
```
I have tried a variety of switches, internal variables and verbosity settings to avoid these, but so far nothing removes them. These are noisy but harmless output, as such I think we have to live with them...

I'm happy to accept this PR, especially as we have other things waiting on CMake builds - @BenBE ?

BenBE:
I think it's not really nice, but not a show-stopper either.

Mostly remarked on this issue to bring it into attention to see if we could have it less noisy, but if that's the way it is I'm fine with it. if we later find another way to achieve the same, we can still update this by another PR. The only thing I'd require is that modified dependencies in TF are properly handled, i.e. when any source file there is touched that it also properly rebuilds everything. There should be no need for anyone ever to manually have to trigger the TF rebuild. As long as that's given this PR is fine by me.

phlash:
Confirmed - touching a random file within TFLite causes re-compile/re-link flow as expected. Merged.

PR #51, opened 2021-02-22
FEATURE: Allow setting video format of input source

oleid:
It would seem some cameras need special attention. I've got a `Logitech BRIO` and it won't allow using higher resolutions, unless I use `MJPG` format as video source. `YUV` only works here for 640x480. And it would seem I need a higher resolution to get the fake webcam to work in my Chromium (88.0.4324.182).

The following patch extends the command line options to include the option `-f`. It accepts a string, which is then turned into an option of `CV_CAP_PROP_FOURCC`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,8 @@ Then, run deepseg (-d -d for full debug, -c for capture device, -v for virtual d
 ./deepseg -d -d -c /dev/video0 -v /dev/video1
 ```
 
+Some cameras like a *Logitec Brio* might need to switch the video source to `MJPG` by passing `-f MJPG` in order to allow for higher resolutions.
+
 ## Limitations/Extensions
 
 As usual: pull requests welcome.
--- a/deepseg.cc
+++ b/deepseg.cc
@@ -15,6 +15,8 @@ limitations under the License.
 #include <unistd.h>
 #include <cstdio>
 #include <chrono>
+#include <string>
+
 #include "tensorflow/lite/interpreter.h"
 #include "tensorflow/lite/kernels/register.h"
 #include "tensorflow/lite/model.h"
@@ -28,6 +30,30 @@ limitations under the License.
 
 #include "loopback.h"
 #include "transpose_conv_bias.h"
+
+int fourCcFromString(const std::string& in)
+{
+	if (in.empty())
+		return 0;
+
+	if (in.size() <= 4)
+	{
+		// fourcc codes are up to 4 bytes long, right-space-padded and upper-case
+		// c.f. http://ffmpeg.org/doxygen/trunk/isom_8c-source.html and
+		// c.f. https://www.fourcc.org/codecs.php
+		std::array<uint8_t, 4> a = {' ', ' ', ' ', ' '};
+		for (auto i = 0; i < in.size(); ++i)
+			a[i] = ::toupper(in[i]);
+		return cv::VideoWriter::fourcc(a[0], a[1], a[2], a[3]);
+	}
+	else if (in.size() == 8)
+	{
+		// Most people seem to agree on 0x47504A4D being the fourcc code of "MJPG", not the literal translation
+		// 0x4D4A5047. This is also what ffmpeg expects.
+		return std::stoi(in, nullptr, 16);
+	}
+	return 0;
+}
 
 // OpenCV helper functions
 cv::Mat convert_rgb_to_yuyv( cv::Mat input ) {
@@ -159,6 +185,7 @@ int main(int argc, char* argv[]) {
 	const char *ccam = "/dev/video1";
 	bool flipHorizontal = false;
 	bool flipVertical   = false;
+	int fourcc = 0;
 
 	const char* modelname = "models/segm_full_v679.tflite";
 
@@ -213,6 +240,15 @@ int main(int argc, char* argv[]) {
 			} else {
 				showUsage = true;
 			}
+		} else if (strncmp(argv[arg], "-f", 2)==0) {
+			if (hasArgument) {
+				fourcc = fourCcFromString(argv[++arg]);
+				if (!fourcc) {
+					showUsage = true;
+				}
+			} else {
+				showUsage = true;
+			}
 		} else if (strncmp(argv[arg], "-t", 2)==0) {
 			if (hasArgument && sscanf(argv[++arg], "%d", &threads)) {
 				if (!threads) {
@@ -236,6 +272,7 @@ int main(int argc, char* argv[]) {
 		fprintf(stderr, "-v            Specify the video target (sink) device\n");
 		fprintf(stderr, "-w            Specify the video stream width\n");
 		fprintf(stderr, "-h            Specify the video stream height\n");
+		fprintf(stderr, "-f            Specify the camera video format, i.e. MJPG or 47504A4D.\n");
 		fprintf(stderr, "-t            Specify the number of threads used for processing\n");
 		fprintf(stderr, "-b            Specify the background image\n");
 		fprintf(stderr, "-m            Specify the TFLite model used for segmentation\n");
@@ -278,6 +315,8 @@ int main(int argc, char* argv[]) {
 
 	cap.set(CV_CAP_PROP_FRAME_WIDTH,  width);
 	cap.set(CV_CAP_PROP_FRAME_HEIGHT, height);
+	if (fourcc)
+		cap.set(CV_CAP_PROP_FOURCC, fourcc);
 	cap.set(CV_CAP_PROP_CONVERT_RGB, true);
 
 	// Load model


BenBE:
Please check if #48 already addresses this issue.

deepseg.cc
```diff
@@ -278,6 +286,8 @@ int main(int argc, char* argv[]) {
 
 	cap.set(CV_CAP_PROP_FRAME_WIDTH,  width);
 	cap.set(CV_CAP_PROP_FRAME_HEIGHT, height);
+	if (fmt.size() > 3)
+		cap.set(CV_CAP_PROP_FOURCC, *(uint32_t*)(fmt.data()));
```
BenBE, line 290: Allow for both 4 bytes (raw) and 8 hex digits variants to be provided. Check should be on these exact lengths.

deepseg.cc
```diff
@@ -344,9 +354,8 @@ int main(int argc, char* argv[]) {
 		// switch buffer pointers in capture thread
 		pthread_mutex_lock(&capinfo.lock);
 		ti.lockns=timestamp();
-		cv::Mat *tmat = capinfo.grab;
-		capinfo.grab = capinfo.raw;
-		capinfo.raw = tmat;
+		std::swap(capinfo.grab, capinfo.raw);
+
```
BenBE, line 358: Please split into a separate cleanup PR. TIA.

README.md
```diff
@@ -130,6 +130,8 @@ Then, run deepseg (-d -d for full debug, -c for capture device, -v for virtual d
 ./deepseg -d -d -c /dev/video0 -v /dev/video1
 ```
 
+Some cameras like a *Logitec Brio* might need to switch the video source to `MJPG` by passing `-f MJPG` in order to reach higher resolutions.
```
BenBE, line 133: reach -> provide/allow for …

BenBE (changes requested):

oleid:
> Please check if #48 already addresses this issue.

It doesn't solve the issue, but it greatly reduced the patch I was preparing :)

My problem is a device timeout. So really a hardware issue.

deepseg.cc
oleid: Oh right, it was "YUYV" not "YUY". I recalled incorrectly.

> Allow for both 4 bytes (raw) and 8 hex digits variants to be provided.

Do you happen to have any documentation on the allowed format?

BenBE:
Looking forward to the updated patch. Basic idea already looks good.

phlash:
Thanks @oleid - if you _don't_ specify a format, but only a resolution, does OpenCV make sensible choices, or assume a format that doesn't work? Asking as it might only be necessary to _force_ OpenCV's hand in rare circumstances, which would make this easier to use for a majority :)

deepseg.cc
phlash: I've been working from the OpenCV docs: https://docs.opencv.org/3.2.0/dd/de7/group__videoio.html - your version may vary as we've discovered. I also find I have to frequently dive into the source: https://github.com/opencv/opencv/tree/3.2.0, and in particular the `videoio` module: https://github.com/opencv/opencv/tree/3.2.0/modules/videoio/src and `cap_v4l.cpp` especially to find out what assumptions it makes, or if it's a leaky abstraction (it often is).

oleid:
> if you _don't_ specify a format, but only a resolution, does OpenCV make sensible choices, or assume a format that doesn't work? 

`v4l2-ctl --list-formats-ext` lists `[0]: 'YUYV' (YUYV 4:2:2)` up to `1920x1080`. OpenCV tries to use that codec automatically.

I'm not sure if OpenCV is to blame, since the camera works in e.g. Chromium or Firefox (I don't know which codec they use, though). But given `YUYV` is reported to work, my best guess is the driver is broken somehow.

deepseg.cc
BenBE: Regarding the two formats: Providing the code as 4 raw bytes might interfere with character encodings or `NUL` bytes in the format code. Thus a plain `uint32_t` might work for most common codes, but it is still advisable to just handle some format that doesn't suffer from this issue and allows for all possible codes to be represented.

So sticking to your example: `MJPG` and `4D4A5047` and `4d4a5047` are identical (note the big endian for the hex).

deepseg.cc
oleid: done

README.md
oleid: done

deepseg.cc
oleid: It would seem there are also 3 byte raw variants and thus can be less-than-8-hex-digit-variants. I hope I've addressed all possibilities in my latest change. Please re-open if not.

BenBE (commented):
Some minor notes:
1. check for `in.size() == 8`. This avoids ambiguity
2. the hex-format part is interpreted in the wrong byte-order. Note my previous example on how to provide MJPG in hex. The current source will fail with that example.

deepseg.cc
oleid: > Some minor notes:
> 
>     1. check for `in.size() == 8`. This avoids ambiguity

My reasoning was: people might skip leading zeros. But I don't have any strong preferences.

>     2. the hex-format part is interpreted in the wrong byte-order. Note my previous example on how to provide MJPG in hex. The current source will fail with that example.

Maybe stupid question: why do you think `MJPG` should be represented as `0x4D4A5047`?

From what I can tell, people[1,2] expect  `0x47504A4D` (which is what you'd get when translating by hand `ord('M')  | ord('J') << 8 | ord('P') << 16 | ord('G') << 24`)

I also saw this calculation in ffmpeg sources (which I linked in the code). 

* [1] https://gix.github.io/media-types/
* [2] https://stackoverflow.com/questions/54740012/error-while-saving-video-as-mp4-using-videowriteropencv-in-ubuntu

oleid:
> Some minor notes:
> 
>     1. check for `in.size() == 8`. This avoids ambiguity

My reasoning was: people might skip leading zeros. But I don't have any strong preferences.

>     2. the hex-format part is interpreted in the wrong byte-order. Note my previous example on how to provide MJPG in hex. The current source will fail with that example.

Maybe stupid question: why do you think `MJPG` should be represented as `0x4D4A5047`?

From what I can tell, people[1,2] expect  `0x47504A4D` (which is what you'd get when translating by hand `ord('M')  | ord('J') << 8 | ord('P') << 16 | ord('G') << 24`)

I also saw this calculation in ffmpeg sources (which I linked in the code). 

* [1] https://gix.github.io/media-types/
* [2] https://stackoverflow.com/questions/54740012/error-while-saving-video-as-mp4-using-videowriteropencv-in-ubuntu

BenBE:
> > 1. the hex-format part is interpreted in the wrong byte-order.
> >    Note my previous example on how to provide MJPG in hex. The current source will fail with that example.
> 
> Maybe stupid question: why do you think `MJPG` should be represented as `0x4D4A5047`?
> 
> From what I can tell, people[1,2] expect `0x47504A4D` (which is what you'd get when translating by hand `ord('M') | ord('J') << 8 | ord('P') << 16 | ord('G') << 24`)
> 
> I also saw this calculation in ffmpeg sources (which I linked in the code).
> 
> * [1] https://gix.github.io/media-types/
> * [2] https://stackoverflow.com/questions/54740012/error-while-saving-video-as-mp4-using-videowriteropencv-in-ubuntu

I came from to that encoding by specifying the forcc byte-wise in just the order the byte values are given - just like a shortened hexdump of it. But if it's documented the other way around I'm fine with that to avoid confusing people. Maybe add a small note with an usage example in the source.

oleid:
> Some minor notes:
> 
>     1. check for `in.size() == 8`. This avoids ambiguity


Done. I also added comments and made sure that the program doesn't run if `fourCcFromString`  doesn't accept the string.

deepseg.cc
```diff
@@ -213,6 +240,15 @@ int main(int argc, char* argv[]) {
 			} else {
 				showUsage = true;
 			}
+		} else if (strncmp(argv[arg], "-f", 2)==0) {
+			if (hasArgument) {
+				fourcc = fourCcFromString(argv[++arg]);
+				if (!fourcc) {
```
BenBE, line 246: Ambiguous regarding `0` vs. invalid. Use `std::optional` to disambiguate.

oleid: I prefer optional myself, however:

* Are we on C++17? I don't see `-std=c++17` set in the Makefile and I'm not sure if Ubuntu 18.04 supports that.
* Zero is really invalid here. A fourcc of 0 doesn't make sense.

oleid: We could use `absl::optional` even with C++11. It is used within tensorflow. That would, however, also require adapting the includes with the `Makefile` based build.

BenBE: Good point regarding the language standard (I'm used to C++17 and newer for my own projects) …

Iff 0 is really undefined/invalid this might be used for now, until the language standard issue is decided. Usually magic numbers should be avoided.

oleid: > Usually magic numbers should be avoided.

I fully agree :)

oleid:
Is there anything missing to get this merged?

phlash:
This all looks good to me, merging.

PR #53, opened 2021-02-24
Add program option to explicitly show the progress bar

cristicc:
Hi,

I added an option `-p` to enable explicitly the display of the dotted progress bar, by default it won't be shown.

The rationale is that, with this feature always enabled, it will pollute the terminal where the program was launched without bringing much value. For monitoring the activity I think `-d` (given one time) is a better approach, anyway.

Additionally, I have updated the usage info to group width & height options under a single optional expression (i.e. `[-w <width> -h <height>]`) since they cannot be used separately.

Thanks for the great work,
Cristi

### DIFF
--- a/deepseg.cc
+++ b/deepseg.cc
@@ -70,7 +70,7 @@ cv::Mat getTensorMat(int tnum, int debug) {
 	TfLiteIntArray* dims = interpreter->tensor(tnum)->dims;
 	if (debug) for (int i = 0; i < dims->size; i++) printf("tensor #%d: %d\n",tnum,dims->data[i]);
 	TFLITE_MINIMAL_CHECK(dims->data[0] == 1);
-	
+
 	int h = dims->data[1];
 	int w = dims->data[2];
 	int c = dims->data[3];
@@ -151,6 +151,7 @@ int main(int argc, char* argv[]) {
 	timinginfo_t ti;
 	ti.bootns = timestamp();
 	int debug  = 0;
+	bool showProgress = false;
 	int threads= 2;
 	int width  = 640;
 	int height = 480;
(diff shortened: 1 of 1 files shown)


deepseg.cc
```diff
@@ -227,11 +230,12 @@ int main(int argc, char* argv[]) {
 	if (showUsage) {
 		fprintf(stderr, "\n");
 		fprintf(stderr, "usage:\n");
-		fprintf(stderr, "  deepseg [-?] [-d] [-c <capture>] [-v <virtual>] [-w <width>] [-h <height>]\n");
+		fprintf(stderr, "  deepseg [-?] [-d] [-p] [-c <capture>] [-v <virtual>] [-w <width> -h <height>]\n");
```
BenBE, line 233: While it doesn't usually make much sense to leave out width or height, but specify the other, it's not forbidden by program logic to do so. And as they have default values just providing one may work just as intended (`-w 800` to get `800x480` as resolution or single `-h 400` to get `640x400` which are both valid resolutions).

Also this is an unrelated change that's better suited for another PR …

BenBE (changes requested):

deepseg.cc
cristicc: Totally agree, I will revert the changes around width and height. Thanks for reviewing!

BenBE (approved):

phlash:
Looks good to me, merging. @cristicc FYI we are looking to refactor the debug/display parts in the [experimental branch](/floe/deepbacksub/tree/experimental), along with more dynamic controls (on/off, flip, resize, change camera devices...) that would be useful in a live presentation situation where we cannot drop and restart the virtual camera, thus we cannot restart this app.

PR #59, opened 2021-04-04
Create thread for mask calculation

peckto:
This is a new implementation of #38 

I moved the grab functionality into the main thread and created a new thread for the mask calculation.

### DIFF
--- a/app/deepseg.cc
+++ b/app/deepseg.cc
@@ -10,6 +10,7 @@
 #include <string>
 #include <thread>
 #include <mutex>
+#include <condition_variable>
 
 #include <opencv2/core/core.hpp>
 #include <opencv2/imgproc/imgproc.hpp>
@@ -133,29 +134,6 @@ typedef struct {
 	timinginfo_t *pti;
 	std::mutex lock;
 } capinfo_t;
-
-// capture thread function
-void grab_thread(capinfo_t *ci) {
-	bool done = false;
-	// while we have a grab frame.. grab frames
-	while (!done) {
-		timestamp_t ts = timestamp();
-		ci->cap->grab();
-		long ns = diffnanosecs(timestamp(),ts);
-		{
-			std::lock_guard<std::mutex> hold(ci->lock);
-			ci->pti->grabns = ns;
-			if (ci->grab!=NULL) {
-				ts = timestamp();
-				ci->cap->retrieve(*ci->grab);
-				ci->pti->retrns = diffnanosecs(timestamp(),ts);
-			} else {
-				done = true;
-			}
-			ci->cnt++;
-		}
-	}
-}
 
 // timing callbacks
 void onprep(void *ctx) { ((timinginfo_t *)ctx)->prepns=timestamp(); }
@@ -356,12 +334,8 @@ int main(int argc, char* argv[]) {
 	if (!maskctx)
 		exit(1);
 
-	// kick off separate grabber thread to keep OpenCV/FFMpeg happy (or it lags badly)
-	cv::Mat buf1;
-	cv::Mat buf2;
-	int64 oldcnt = 0;
-	capinfo_t capinfo = { &cap, &buf1, &buf2, 0, &ti };
-	std::thread grabber(grab_thread, &capinfo);
+	cv::Mat mask;
+	cv::Mat raw;
 	ti.lastns = timestamp();
 	printf("Startup: %ldns\n", diffnanosecs(ti.lastns,ti.bootns));
 
@@ -369,21 +343,13 @@ int main(int argc, char* argv[]) {
 
 	// mainloop
 	for(bool running = true; running; ) {
-		// wait for next frame
-		while (capinfo.cnt == oldcnt) usleep(10000);
-		oldcnt = capinfo.cnt;
 		ti.waitns=timestamp();
 
-		// switch buffer pointers in capture thread
-		{
-			std::lock_guard<std::mutex> hold(capinfo.lock);
-			ti.lockns=timestamp();
-			cv::Mat *tmat = capinfo.grab;
-			capinfo.grab = capinfo.raw;
-			capinfo.raw = tmat;
-		}
-		// we can now guarantee capinfo.raw will remain unchanged while we process it..
-		cv::Mat raw = *capinfo.raw;
+		// grab new frame from cam
+		cap.grab();
+		// copy new frame to buffer
+		cap.retrieve(raw);
+
 		ti.copyns=timestamp();
 		if (raw.rows == 0 || raw.cols == 0) continue; // sanity check
 
@@ -474,12 +440,6 @@ int main(int argc, char* argv[]) {
 				break;
 		}
 	}
-
-	{
-		std::lock_guard<std::mutex> hold(capinfo.lock);
-		capinfo.grab = NULL;
-	}
-	grabber.join();
 
 	printf("\n");
 	return 0;
--- a/app/deepseg.cc
+++ b/app/deepseg.cc
@@ -105,17 +105,14 @@ typedef std::chrono::high_resolution_clock::time_point timestamp_t;
 typedef struct {
 	timestamp_t bootns;
 	timestamp_t lastns;
-	timestamp_t waitns;
 	timestamp_t lockns;
 	timestamp_t copyns;
 	timestamp_t prepns;
-	timestamp_t tfltns;
 	timestamp_t maskns;
 	timestamp_t postns;
 	timestamp_t v4l2ns;
-	// these are already converted to ns
-	long grabns;
-	long retrns;
+	timestamp_t grabns;
+	timestamp_t retrns;
 } timinginfo_t;
 
 timestamp_t timestamp() {
@@ -134,11 +131,6 @@ typedef struct {
 	timinginfo_t *pti;
 	std::mutex lock;
 } capinfo_t;
-
-// timing callbacks
-void onprep(void *ctx) { ((timinginfo_t *)ctx)->prepns=timestamp(); }
-void oninfer(void *ctx) { ((timinginfo_t *)ctx)->tfltns=timestamp(); }
-void onmask(void *ctx) { ((timinginfo_t *)ctx)->maskns=timestamp(); }
 
 // let's do this!
 static bool is_number(const std::string &s) {
@@ -170,11 +162,13 @@ protected:
 
 	void run() {
 		cv::Mat *raw_tmp;
+		timestamp_t t1;
 
 		while(thread_state::INIT == this->state)
 			usleep(1000); // Wait for constructor to complete initialization
 
 		while(thread_state::RUNNING == this->state) {
+			t0 = timestamp();
 			/* actual handling */
 			{
 				std::unique_lock<std::mutex> hold(lock_frame);
@@ -188,6 +182,9 @@ protected:
 				frame_next = frame_current;
 				frame_current = raw_tmp;
 			}
+			waitns=diffnanosecs(timestamp(), t0);
+			t0 = timestamp();
+			t1 = timestamp();
 			if(!bs_maskgen_process(maskctx, *frame_current, *mask_current)) {
 				fprintf(stderr, "failed to process video frame\n");
 				exit(1);
@@ -199,15 +196,56 @@ protected:
 				mask_current = raw_tmp;
 				new_mask = true;
 			}
+			loopns = diffnanosecs(timestamp(), t1);
 		}
 	}
 
+	// timing callbacks
+	static void onprep(void *ctx) {
+		CalcMask *cls = (CalcMask *)ctx;
+		cls->prepns=diffnanosecs(timestamp(), cls->t0);
+		cls->t0 = timestamp();
+	}
+	static void oninfer(void *ctx) {
+		CalcMask *cls = (CalcMask *)ctx;
+		cls->tfltns=diffnanosecs(timestamp(), cls->t0);
+		cls->t0 = timestamp();
+	}
+	static void onmask(void *ctx) {
+		CalcMask *cls = (CalcMask *)ctx;
+		cls->maskns=diffnanosecs(timestamp(), cls->t0);
+		cls->t0 = timestamp();
+	}
+
 public:
-	CalcMask(void *maskctx) :
+	timestamp_t t0;
+	long waitns;
+	long prepns;
+	long tfltns;
+	long maskns;
+	long loopns;
+
+	CalcMask(const char *modelname,
+			 size_t threads,
+			 size_t width,
+			 size_t height) :
 		state{thread_state::INIT},
 		thread{&CalcMask::run, this} {
+		maskctx = bs_maskgen_new(
+					modelname,
+					threads,
+					width,
+					height,
+					nullptr,
+					onprep,
+					oninfer,
+					onmask,
+					this
+					);
+		if (!maskctx)
+			exit(1);
+
 		// Do all other initialization …
-		this->maskctx = maskctx;
 		frame_next = &frame1;
 		frame_current = &frame2;
 		mask_current = &mask1;
@@ -220,6 +258,7 @@ public:
 	~CalcMask() {
 		state = thread_state::DONE;
 		thread.join();
+		bs_maskgen_delete(maskctx);
 	}
 
 	void set_input_frame(cv::Mat &frame) {
@@ -423,13 +462,9 @@ int main(int argc, char* argv[]) {
 		cap.set(CV_CAP_PROP_FOURCC, fourcc);
 	cap.set(CV_CAP_PROP_CONVERT_RGB, true);
 
-	void *maskctx = bs_maskgen_new(modelname, threads, width, height, nullptr, onprep, oninfer, onmask, &ti);
-	if (!maskctx)
-		exit(1);
-
 	cv::Mat mask(height, width, CV_8U);
 	cv::Mat raw;
-	CalcMask ai(maskctx);
+	CalcMask ai(modelname, threads, width, height);
 	ti.lastns = timestamp();
 	printf("Startup: %ldns\n", diffnanosecs(ti.lastns,ti.bootns));
 
@@ -437,21 +472,22 @@ int main(int argc, char* argv[]) {
 
 	// mainloop
 	for(bool running = true; running; ) {
-		ti.waitns=timestamp();
-
 		// grab new frame from cam
 		cap.grab();
+		ti.grabns=timestamp();
 		// copy new frame to buffer
 		cap.retrieve(raw);
+		ti.retrns=timestamp();
 		ai.set_input_frame(raw);
-
 		ti.copyns=timestamp();
+
 		if (raw.rows == 0 || raw.cols == 0) continue; // sanity check
 
 		if (blur_strength) {
 			raw.copyTo(bg);
 			cv::GaussianBlur(bg,bg,cv::Size(blur_strength,blur_strength),0);
 		}
+		ti.prepns = timestamp();
 
 		if (filterActive) {
 			// do background detection magic
@@ -459,10 +495,8 @@ int main(int argc, char* argv[]) {
 
 			// alpha blend background over foreground using mask
 			raw = alpha_blend(bg, raw, mask);
-		} else {
-			// fix up timing values
-			ti.maskns=ti.tfltns=ti.prepns=ti.copyns;
 		}
+		ti.maskns = timestamp();
 
 		if (flipHorizontal && flipVertical) {
 			cv::flip(raw,raw,-1);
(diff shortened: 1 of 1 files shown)


BenBE:
IMHO it's an advantage to have the grabber thread and the mask calculation in different threads. Ideally, and this is somewhat suboptimal right now, the grabber thread just provides images it captured from the camera while the mask calculation thread latches onto the last grabbed frame whenever it is ready for the next frame for processing. That way frames could be silently discarded when processing a mask is too slow.

phlash:
My fork takes a similar approach to that suggested by Ben: a grab-mask-output thread that runs at full video frame rate using only OpenCV and V4L2, into which the mask calculator thread inserts an updated mask, generated at whatever rate it can maintain, from the latest frame grabbed. This is also the approach taken by [ViBa](https://github.com/fisadev/virtualbackground), and Jitsi to minimise video lag, even if the ML is a bit challenged on cycles.

Note that the grab thread at present exists to ensure we read input frames promptly, again to reduce lag, as OpenCV will queue up video rather than discard it (and there are few/no controls over this). As long as we continue to read-mask-write quickly I don't mind which thread has which responsibility :)

BenBE:
The approach with updating the mask async and passing frames thru with a maybe slightly outdated mask seems fine too for most situations. Might be an option to either discard frames or use the old mask.

phlash:
@peckto I have now read through your changes in detail, and tried them out locally, it looks good. I'll mark up a couple of syntax things I spotted, but otherwise great stuff!

Do we think anyone would want this to operate synchronously below the input frame rate (ie: old behaviour)? If so, how?

deepseg.cc
```diff
@@ -225,11 +202,23 @@ void init_tensorflow(calcinfo_t &info) {
...
 }
 
-void calc_mask(calcinfo_t &info, timinginfo_t &ti) {
+void calc_mask(calcinfo_t &info) {
+	timestamp_t t0 = timestamp();
+
+	// copy frame from shared buffer
+	cv::Mat raw;
+	pthread_mutex_lock(&info.lock_raw);
+	raw = info.raw.clone();
```
phlash, line 217: Can we avoid the double cloning (in/out of shared buffer)? Unsure of performance impact...

deepseg.cc
```diff
@@ -298,15 +289,26 @@ void calc_mask(calcinfo_t &info, timinginfo_t &ti) {
...
 	// denoise
 	cv::Mat tmpbuf;
 	cv::dilate(info.ofinal,tmpbuf,info.element);
 	cv::erode(tmpbuf,info.ofinal,info.element);
 
 	// scale up into full-sized mask
-	cv::resize(info.ofinal,info.mroi,cv::Size(info.raw.rows/info.ratio,info.raw.rows));
+	pthread_mutex_lock(&info.lock_mask);
+	cv::resize(info.ofinal,info.mroi,cv::Size(raw.rows/info.ratio,raw.rows));
+	pthread_mutex_unlock(&info.lock_mask);
```
phlash, line 302: Perhaps calculate a separate frame rate for mask generation (and display in stats)?

deepseg.cc
```diff
@@ -298,15 +289,26 @@ void calc_mask(calcinfo_t &info, timinginfo_t &ti) {
...
+}
+
+void *calc_mask_thread(void *args) {
+	calcinfo_t *calcinfo = (calcinfo_t *)args;
```
phlash, line 306: Would we do better to use a `pthread_cond_wait()/pthread_cond_signal()` pair below to ensure mask calculation only occurs at <= video frame rate? Otherwise we guarantee to consume a core (or two!) calculating flat out. Of course the `start` flag needs to remain (possibly renamed `running`) so we can terminate the thread.

deepseg.cc
BenBE: Structurally we could pre-allocate a few picture buffers and just hand out the pointers to them. This would require proper tracking which pointers are in use at any time and would require some preparations not yet present in the current code.

deepseg.cc
```diff
@@ -464,65 +466,71 @@ int main(int argc, char* argv[]) {
...
+		// copy frame to ai thread local buffer
+		pthread_mutex_lock(&calcinfo.lock_raw);
 		ti.lockns=timestamp();
-		cv::Mat *tmat = capinfo.grab;
-		capinfo.grab = capinfo.raw;
-		capinfo.raw = tmat;
-		pthread_mutex_unlock(&capinfo.lock);
-		// we can now guarantee capinfo.raw will remain unchanged while we process it..
-		calcinfo.raw = *capinfo.raw;
+		calcinfo.raw = raw.clone();
```
phlash, line 504: see above comment regarding double clone, we may want to avoid this one, if there is any material performance improvement.

deepseg.cc
```diff
@@ -464,65 +466,71 @@ int main(int argc, char* argv[]) {
...
+		pthread_mutex_unlock(&calcinfo.lock_raw);
 		ti.copyns=timestamp();
-		if (calcinfo.raw.rows == 0 || calcinfo.raw.cols == 0) continue; // sanity check
+
+		// signal ai thread data arrived for the first time
+		calcinfo.start = true;
```
phlash, line 509: see above regarding `pthread_cond_signal()` here

deepseg.cc
```diff
@@ -464,65 +466,71 @@ int main(int argc, char* argv[]) {
...
 		if (filterActive) {
-			// do background detection magic
-			calc_mask(calcinfo, ti);
+			pthread_mutex_lock(&calcinfo.lock_mask);
+			mask = calcinfo.mask.clone();
+			pthread_mutex_unlock(&calcinfo.lock_mask);
 
 			// copy background over raw cam image using mask
-			bg.copyTo(calcinfo.raw,calcinfo.mask);
+			bg.copyTo(raw,calcinfo.mask);
```
phlash, line 517: BUG: should be using local clone of `mask`

deepseg.cc
peckto: Thanks for the hint!
I implemented it. Is this the way you were thinking of?

deepseg.cc
peckto: done

deepseg.cc
peckto: fixed

peckto:
Regarding opencv and the grab thread:
If I understood you correctly, the main problem is, that opencv does buffer frames, which can lead to a delay if we do not grad the new frame with the maximum frame rate.
In the current (1-thread) implementation, this would occure only, if the clipping and v4l2loopback step would take too long. 
I would assume, that this is under normal conditions not the case. Additionally, opencv now provides a buffer setting (`CAP_PROP_BUFFERSIZE`) for the V4L backend:
https://github.com/opencv/opencv/blob/master/modules/videoio/src/cap_v4l.cpp

With this, I think we can work with the single thread option.
An extra grab thread makes things quite complicated in my opinion and the profit is probably not so high.

peckto:
@phlash: Thank you for your feedback! I will address your points the next days. 

I'm just not sure, how we should proceed with the "double copy" thing.
If this is not really a performance issue, I prefer to keep it simple.
(The copy performance is of course dependent on the video resolution.)

BenBE:
It's preferable to avoid any large copies of memory buffers. Both for performance reasons and for reasons of memory/resource usage. An easy solution is to pre-allocate several buffers large enough to hold a frame and just hand pointers to these allocated objects around. With the current architecture of one thread for grabbing, one for mask generation and the main thread handling the final frame merging this means we'd be fine with a pool of about 8 buffers (one active and one inactive per thread, plus some small reserve). That way each thread could hold on to the last active buffer it acquired from the previous step each without blocking and still allow for enough slack that slow mask calculations won't block everything on the main thread too much, all ensuring constant memory usage..

deepseg.cc
BenBE: As we are using C++ I'd even suggest using the thread toolkit available with C++11 and newer. This even allows for implementing proper RAII paradigms which strongly avoids issues with those primitives not appearing pairwise.

phlash: Heh - you can tell I'm too old for modern C++ :) good shout @BenBE

phlash:
I'd like to hold off on a buffer pool until we know the impact of a couple of clone() calls between threads - the timing we have should make it obvious if we have a problem or not. Thanks for the speedy work @peckto!

deepseg.cc
peckto: Good Idea
Done

peckto:
I added some more measurements for the ai thread.
On my machine it looks like the following:
```
$ ./deepseg -d -d -c /dev/video0 -v /dev/video20 -H 
video [grab: 58741363 retr:   721098 lock:      471 copy:   632780 post:   928098 v4l2:  2864112] ai [cond: 30599240 copy:   183267 pre:  1087840 tflt: 31788445 post:   680549 mask:   144196] - FPS: 16.56 (video) 19.19 (ai)
video [grab: 54829101 retr:   452109 lock:      365 copy:   314983 post:   950727 v4l2:  3267526] ai [cond: 26038497 copy:   230877 pre:   995302 tflt: 31148558 post:   778711 mask:   173230] - FPS: 18.12 (video) 15.63 (ai)
video [grab: 54883606 retr:  1623516 lock:      521 copy:   744499 post:   959518 v4l2:  3415035] ai [cond: 25964401 copy:   180704 pre:  1125331 tflt: 33466405 post:   686013 mask:   155247] - FPS: 17.22 (video) 16.24 (ai)
video [grab: 53589089 retr:   500830 lock:     1068 copy:   354910 post:   925405 v4l2:  3216689] ai [cond: 23785185 copy:   163880 pre:  1027631 tflt: 32905018 post:   707593 mask:   149251] - FPS: 18.61 (video) 16.38 (ai)
video [grab: 54525152 retr:  1019776 lock:      653 copy:   974318 post:  1064244 v4l2:  3157460] ai [cond: 21241901 copy:   244925 pre:  1312905 tflt: 37411994 post:   690571 mask:   179552] - FPS: 17.48 (video) 15.81 (ai)
video [grab: 53905385 retr:   452957 lock:      835 copy:   319837 post:   968303 v4l2:  3224764] ai [cond: 25224082 copy:   239520 pre:   966928 tflt: 31087658 post:   702594 mask:   343650] - FPS: 18.21 (video) 18.20 (ai)
```
(For some reason I currently don't get 30fps??)

I agree with you, that in general we should avoid double copy. 
Maybe we could re-use the double buffer structure from the grabber thread?

phlash:
@peckto, @BenBE: I'm not sure where we are with this PR? To me it looks like @peckto is waiting on a question about double buffering, and then we'll want to check performance is ok before we merge? We can address a switch to the C++11 thread toolkit in a separate PR (as per experimental branch).

Regarding the double buffering - I think that should work, with the 'slower' thread (always mask calculation) swapping buffer pointers (while holding a lock) as it either consumes or writes them. This avoids buffer pool management, while also avoiding multiple copying of video frames.

deepseg.cc
BenBE: Nah, not everybody needs to be fully up-to-date on everything all the time. But it sure helps to have a look around once in a while. ;-)

BenBE:
Haven't tested this yet and with the larger structural rework I think having this land on experimental first is preferable (@floe  What do you think?). This also has the advantage of already having the C++11 patches as a basis to work from.

Haven't had too much time for thorough reviews and tests of the experimental branch the last days, so I'd recommend the experimental changes ripen there for another few days before we merge them to main.

Will try to get a review for this PR tomorrow or on Monday as time permits.

peckto:
I added the double buffer structure for the frame buffer Mat. The main thread copies the received frame to the `next` buffer, while the ai thread works on the `current` buffer and changing the buffers on every cycle. In this way, we have only one copy operation for the frame.

peckto:
I had a misunderstanding regarding the `pthread_cond_signal` and `pthread_cond_wait`. I thought, the wait would be async, in a way, that the condition is always true, as long as the signal arrived before the last wait. But as far as I can see, the wait is only released, after the next signal. This caused a unnecessary long wait period at the `pthread_cond_wait` call. To fix this issue, I new use a flag, to indicate that a new frame has arrived and the ai thread just sleeps until the flags gets true.
I'm not so much into threading. If you know a cleaner way to achieve this, please let me know.

BenBE:
You might be looking for this:
https://en.cppreference.com/w/cpp/thread/condition_variable/wait_for

phlash:
@peckto - yep you are correct, I re-read the manual (https://linux.die.net/man/3/pthread_cond_wait):
> When using condition variables there is always a Boolean predicate involving shared variables associated with each condition wait that is true if the thread should proceed.

So there should indeed be a new frame indicator, set just before raising the signal, checked/cleared after wakeup in the other thread eg:
```c
    // In calc_mask()
    // ...
    pthread_mutex_lock(&info.lock_raw);
    while (!info.new_frame) {
        pthread_cond_wait(&info.condition_new_frame, &info.lock_raw);
    }
    info.new_frame = false;
    // ...any other stuff requiring the lock
    pthread_mutex_unlock(&info.lock_raw);
...
    // In main()
    // ...
    pthread_mutex_lock(&calcinfo.lock_raw);
    // ...any other stuff requiring the lock
    calcinfo.new_frame = true;
    pthread_cond_signal(&calcinfo.condition_new_frame);
    pthread_mutex_unlock(&calcinfo.lock_raw);
```
this will avoid the issue you noticed where the mask calculation rate drops to half frame rate, since it's waiting for the next signal even when a new frame is already available.
As @BenBE linked to, the C++11 thread library works in a similar manner (although it has a helper function that does the predicate wait loop for you)

peckto:
Combining `pthread_cond_wait` and the flag is a good idea. I implemented your suggestion. 
I noticed, the experimental branch now has patches to adapt to the C++ threading library.
I'm not quite sure, if I understood the order in which you want to merge things into main.

BenBE:
The major refactoring work normally should to to experimental, main should be left for minor fixups and continuous improvements, that are unlikely to cause stability issues. Cf. [here](https://github.com/floe/deepbacksub/pull/59#issuecomment-821901028). Thus this should probably be going to experimental too, as note above.

phlash:
@peckto, @BenBE  polite nudge for this one - I suggest rebasing to `experimental`, and ensuring C++11 thread library is used? then we can pull this in, as I find it makes a material improvement.

BenBE:
> I suggest rebasing to `experimental`, and ensuring C++11 thread library is used? then we can pull this in, as I find it makes a material improvement.

ACK.

peckto:
I'm working on the rebase to the experimental branch.

I noticed the PR #86, which seems to have overlapping code changes with this PR.
Do you have a plan, in which order you want to merge these two PRs to minimize merge conflicts?

BenBE:
@peckto Any preference on your side? IMHO splitting things into a library first and afterwards extracting some operations into their own thread is probably the more efficient way to go, as the library split forces a separation of concerns in some sense. What do you think?

phlash:
I think #86 is going to need a fair amount of tidying up done before it gets merged and I'd like this one in sooner to improve UX. IMO any threading should be external to TF/the library (which aims to do one job well), so there shouldn't be a clash (although I expect a few function name changes to come along with the separation, they should be trivial fixes).

peckto:
The MR is now rebased to experimental branch and commits are squashed.
For threading, I'm now using the c++ std module.

deepseg.cc
```diff
@@ -371,20 +379,35 @@ void calc_mask(calcinfo_t &info, timinginfo_t &ti) {
...
+	info.pti->ai_totalns = diffnanosecs(timestamp(), t);
 }
 
 bool is_number(const std::string &s) {
   return !s.empty() && std::all_of(s.begin(), s.end(), ::isdigit);
 }
 
+void *calc_mask_thread(void *args) {
+	calcinfo_t *calcinfo = (calcinfo_t *)args;
+	while (!calcinfo->running) usleep(10000);
```
BenBE, line 404: Please avoid having code on the same line as the beginning of a block statement.
Feel free to skip braces for single statements though.

deepseg.cc
```diff
@@ -371,20 +379,35 @@ void calc_mask(calcinfo_t &info, timinginfo_t &ti) {
...
-	cv::blur(tmpbuf,info.mroi,info.blur);
+	info.pti->ai_maskns = diffnanosecs(timestamp(), t0);
+	info.pti->ai_totalns = diffnanosecs(timestamp(), t);
 }
 
 bool is_number(const std::string &s) {
   return !s.empty() && std::all_of(s.begin(), s.end(), ::isdigit);
 }
 
+void *calc_mask_thread(void *args) {
```
BenBE, line 402: What about having this as a proper `thread`-derived class?

deepseg.cc
peckto: fixed

deepseg.cc
peckto: Do you mean something like:
```c++
class CalcMask : public std::thread
```
I'm not so familiar with the threading module. 
Is this the recommended usage?

BenBE: Reading up on the [docs](https://en.cppreference.com/w/cpp/thread/thread) directly inheriting may cause some issues with initialization order (due to the thread being possibly running before the constructor completes). The following way should be more stable against that scenario:

```cpp
class CalcMask final {
protected:
    enum class thread_state { INIT, RUNNING, DONE };
    volatile thread_state state;
    // All mutexes need to come before the thread
    std::thread thread; // Must be after state and all relevant mutexes, because of initialization order)
    void run() {
        while(thread_state::INIT == this->state)
            ; // Wait for constructor to complete initialization
        while(thread_state::RUNNING == this->state) {
            /* actual handling */
        }
    }

public:
    CalcMask() :
        state{thread_state::INIT},
        thread{this->run} {
        // Do all other initialization …
        state = thread_state::RUNNING;
    }

    ~CalcMask() {
        state = thread_state::DONE;
        thread.join();
    }
};
```

peckto: I'm not sure, where the new `CalcMask` class will bring us. 
From a naive perspective, this looks unnecessary complex to me.
Where do you see the benefit of having the thread in a separate class?
What do you see inside the thread class? Only the `calc_mask_thread` method, or the whole mask calculation?

BenBE: The mask calculation + mutexes + thread-local data + functions to interact/synchronize with it.
In the first iteration mostly just the `calc_mask_thread` function + mutexes.

peckto: If I understand your correctly, you want to move the whole mask calculation into a class?
At the moment the `calcinfo_t` struct is a container for data. Isn't creating a `CalcMask` class the same as extending the c like struct into a c++ class?
For me this looks like a larger restructure, including creating a .cpp and .h file for the new class. 
Do you think we should do this as part of this PR?

On the other hand, I'm not so sure if having a class containing just the thread context makes things nicer than the current implementation. Having the thread context, mutex and run loop at one place, but the function called by the run loop and the data associated with the mutex at a different place doesn't sound ideal to me.

What do you think?

BenBE: There's a subtle difference between those two structures which mostly affects how things can be interacted with from code that is not C++. See #87 for issues and challenges with that one.

So the basic idea should be to keep the publicly visible part of the structure as simple as possible, and especially try to avoid C++-isms there, while the internal `CalcMask` object can be as complicated as we need it to be. Think of it kinda as `calcinfo_t` as the parameters and `CalcMask` as the runtime data.

BenBE:
@peckto Do you mind to rebase to resolve the merge conflicts?

peckto:
Hi @BenBE, sorry for the delay. 
I had not much time to work on this project the last weeks.

A rebase turns out to be not straight forward.
Probably a re-write is better. The thread class is missing anyhow.
I'll have a look the next days.

BenBE:
Sure, don't worry. Absolutely no need to rush anything.

deepseg.cc
peckto: I see the point, that you want to have the lib as pure C.
As we now have the split, I would see the threading only as part of the app.

Just for me to get a better overall picture, what is your plan for the app? 
Do you want to restructure the whole main function into a class?
Will this be all part of the above mentioned `CalcMask` class, or will there be more then one class?

BenBE: The `CalcMask` class is likely only one of many classes, centered around the aspect of getting the segmentation mask from an input frame. Other classes/modules may perform tasks like frame rate decoupling (re-use mask for multiple frames, until next mask has been calculated) or image post-processing (hologram effects, light boxing, …).

peckto: Thank you for the explanation. It's easier for me, if I can see the big picture.

peckto:
Having the actual mask calculation in the lib, makes this MR much cleaner I think.
I pushed a first draft of the `CalcMask` class, implementing the threading.
From an architecture point of view, is this the way you were thinking of?

phlash:
Thanks @peckto, that's looking really tidy!

peckto:
I adapted the logging. 
Therefore I moved the `maskctx`and the callbacks into the `CalcMask` class.
Performance looks good on my system.
What do you think?

app/deepseg.cc
```diff
@@ -199,15 +196,56 @@ class CalcMask final {
...
+					modelname,
+					threads,
+					width,
+					height,
+					nullptr,
+					onprep,
+					oninfer,
+					onmask,
+					this
+					);
```
BenBE, line 244: Avoid stretching things onto too many LOC …

app/deepseg.cc
```diff
@@ -199,15 +196,56 @@ class CalcMask final {
...
+		if (!maskctx)
+			exit(1);
```
BenBE, line 246: Shouldn't call exit from non-obvious places. Prefer to raise an exception instead and handle this globally.

app/deepseg.cc
```diff
@@ -495,18 +529,21 @@ int main(int argc, char* argv[]) {
...
+			diffnanosecs(ti.retrns,ti.grabns),
+			diffnanosecs(ti.copyns,ti.retrns),
 			diffnanosecs(ti.prepns,ti.copyns),
-			diffnanosecs(ti.tfltns,ti.prepns),
-			diffnanosecs(ti.maskns,ti.tfltns),
+			diffnanosecs(ti.maskns,ti.prepns),
 			diffnanosecs(ti.postns,ti.maskns),
 			diffnanosecs(ti.v4l2ns,ti.postns),
-			1e9/diffnanosecs(ti.v4l2ns,ti.lastns));
+			1e9/diffnanosecs(ti.v4l2ns,ti.lastns),
```
BenBE, line 540: Whitespace after comma should be homogeneous.

app/deepseg.cc
peckto: That was a leftover, i fixed it.

app/deepseg.cc
peckto: I'm now using an exception.

app/deepseg.cc
peckto: Thanks, I missed that one.

app/deepseg.cc
```diff
@@ -464,7 +454,13 @@ int main(int argc, char* argv[]) {
 
 	cv::Mat mask(height, width, CV_8U);
 	cv::Mat raw;
-	CalcMask ai(modelname, threads, width, height);
+	CalcMask *ai;
+	try {
+		ai = new CalcMask(modelname, threads, width, height);
```
BenBE, line 459: Use `std::unique_ptr<>`/`std::make_unique` or `std::shared_ptr`/`std::make_shared`. Also saves on manual resource management on exit.

BenBE: And why change `ai` from stack object to pointer?

peckto: Because of the exception in the constructor.

BenBE: This shouldn't make a difference, when you stretch the `try`/`catch` over the whole lifetime of `ai` required by the code below. Also inside main, you can just `return 1;` whcih takes care of all the automatic dtor calls required.

peckto: The lifetime of `ai` covers the whole main loop.
This seems like a large block for a try.
Do you think, its better to wrap the main function itself with a try instead?

BenBE: You mean like:

```cpp
int main() try {
    // Code
} catch(const char* msg) {
    fprintf(stderr, "Error: %s\n", msg);
    return 1;
}
```

Fine by me … ;-)

peckto: yes, fixed it.

deepseg.cc
phlash: Looks resolved :smile:

app/deepseg.cc
phlash: Looks resolved :smile:

phlash:
This LGTM, works as expected, merging!

PR #60, opened 2021-04-04
Blur video background

peckto:
Currently the background is replaced by a green screen in case no background image is given.
I thought it would be nice to blur the background of the video stream instead.

Do you have any special use case for the green screen mode?
In this case I can add an extra cli argument for the blur.

### DIFF
--- a/deepseg.cc
+++ b/deepseg.cc
@@ -309,6 +309,10 @@ void calc_mask(calcinfo_t &info, timinginfo_t &ti) {
 	cv::resize(info.ofinal,info.mroi,cv::Size(info.raw.rows/info.ratio,info.raw.rows));
 }
 
+bool is_number(const std::string &s) {
+  return !s.empty() && std::all_of(s.begin(), s.end(), ::isdigit);
+}
+
 int main(int argc, char* argv[]) {
 
 	printf("deepseg v0.2.0\n");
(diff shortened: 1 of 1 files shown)


BenBE:
The green screen mode is essentially useful in combination with other software like OBS Studio for further processing as you can use OBS to filter that solid green background and replace it by more complex compositions that deepbacksub itself does not implement.

peckto:
Ok, I see.
I changed the PR to add an extra blur option `-u`.

`-b` is used for the background image. 
Maybe its better to rename this to `-i` and use `-b` for the blur?

BenBE:
What about `-p bgblur:<strength>` (`-p` for "post-processing"), so we could have others.
Also an idea I had was to blur the mask a bit when copying the image over, so the edges of the selected image part aren't that sharp. The radius for this could similarly be set with something like `-p edgeblur:<strength>`.

BenBE:
Moving this to experimental for now, so we can experiment a bit more with this feature.

peckto:
The post-processing option sounds nice. I also thought about an option to dilate the mask, to optimize the clipping result.

I changed the PR to add the ` -p bgblur:<strength>` option.
Unfortunately there is already the `-p` "Show progress bar" option. I renamed this to `-s` for now.

peckto:
I noticed, the experimental branch is behind master. Do you plan to rebase?

BenBE:
Will have to take a look, but IIRC there's nothing special preventing to rebase experimental ATM. Will do in a moment.

phlash:
@peckto, @BenBE polite nudge for this PR, anything we want done before merging to `experimental` ?

BenBE:
@phlash LGTM.

PR #64, opened 2021-04-15
Switch to C++11 mutexes

phlash:
As per the title, these changes:
 * replace `pthreads` with `std::mutex` classes
 * place output files in a `bin` folder (simplifying .gitignore and clean target)
 * use the current git tag or branch to provide a version string

### DIFF
--- a/deepseg.cc
+++ b/deepseg.cc
@@ -8,6 +8,8 @@
 #include <cstdio>
 #include <chrono>
 #include <string>
+#include <thread>
+#include <mutex>
 
 #include "tensorflow/lite/interpreter.h"
 #include "tensorflow/lite/kernels/register.h"
@@ -113,7 +115,7 @@ typedef struct {
 	timestamp_t waitns;
 	timestamp_t lockns;
 	timestamp_t copyns;
-	timestamp_t openns;
+	timestamp_t prepns;
 	timestamp_t tfltns;
 	timestamp_t maskns;
 	timestamp_t postns;
@@ -137,7 +139,7 @@ typedef struct {
 	cv::Mat *raw;
 	int64 cnt;
 	timinginfo_t *pti;
-	pthread_mutex_t lock;
+	std::mutex lock;
 } capinfo_t;
 
 typedef struct {
@@ -159,15 +161,14 @@ typedef struct {
 } calcinfo_t;
 
 // capture thread function
-void *grab_thread(void *arg) {
-	capinfo_t *ci = (capinfo_t *)arg;
+void grab_thread(capinfo_t *ci) {
 	bool done = false;
 	// while we have a grab frame.. grab frames
 	while (!done) {
 		timestamp_t ts = timestamp();
 		ci->cap->grab();
 		long ns = diffnanosecs(timestamp(),ts);
-		pthread_mutex_lock(&ci->lock);
+		ci->lock.lock();
 		ci->pti->grabns = ns;
 		if (ci->grab!=NULL) {
 			ts = timestamp();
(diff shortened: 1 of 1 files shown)


deepseg.cc
```diff
@@ -159,15 +166,14 @@ typedef struct {
...
-	capinfo_t *ci = (capinfo_t *)arg;
+void grab_thread(capinfo_t *ci) {
 	bool done = false;
 	// while we have a grab frame.. grab frames
 	while (!done) {
 		timestamp_t ts = timestamp();
 		ci->cap->grab();
 		long ns = diffnanosecs(timestamp(),ts);
-		pthread_mutex_lock(&ci->lock);
+		ci->lock.lock();
```
BenBE, line 176: Cf. [`std::lock_guard<Mutex>`](https://en.cppreference.com/w/cpp/thread/lock_guard/lock_guard).

deepseg.cc
```diff
@@ -478,16 +479,15 @@ int main(int argc, char* argv[]) {
...
 
 		// switch buffer pointers in capture thread
-		pthread_mutex_lock(&capinfo.lock);
+		capinfo.lock.lock();
 		ti.lockns=timestamp();
 		cv::Mat *tmat = capinfo.grab;
 		capinfo.grab = capinfo.raw;
 		capinfo.raw = tmat;
-		pthread_mutex_unlock(&capinfo.lock);
+		capinfo.lock.unlock();
```
BenBE, line 490: dito, use braces to denote region for lock to be held.

deepseg.cc
```diff
@@ -569,9 +569,10 @@ int main(int argc, char* argv[]) {
 		}
 	}
 
-	pthread_mutex_lock(&capinfo.lock);
+	capinfo.lock.lock();
 	capinfo.grab = NULL;
-	pthread_mutex_unlock(&capinfo.lock);
+	capinfo.lock.unlock();
```
BenBE, line 574: dito.

Makefile
```diff
@@ -1,13 +1,19 @@
 # this is licensed software, @see LICENSE file.
+
+# OpenCV & Tensorflow recommended flags for performance..
 CFLAGS = -Ofast -march=native -fno-trapping-math -fassociative-math -funsafe-math-optimizations -Wall -pthread
 LDFLAGS = -lrt -ldl
 
+# Version
+VERSION=$(shell git describe --all --always --dirty)
```
BenBE, line 8: Why not `git describe --tags --dirty`? Gives much more descriptive version strings IMHO.

BenBE (changes requested):

BenBE:
Maybe split build version display into its own commit too.

Makefile
phlash: I tried that, but without tags (or with tags in much older shared history) it's less useful (no branch indication, possibly confusing old tag base). This will display a tag if available in the current branch, or the branch itself. We could add `--long` to always show the commit id and revision count from the branch/tag start?

phlash:
OK, will push out the build changes to a new feature branch.

deepseg.cc
phlash: done

deepseg.cc
phlash: done

deepseg.cc
phlash: done

Makefile
phlash: Moved to #68,

BenBE (approved):
LGTM. 

It'd also be sensible to just squash/fixup those two commits into one. I already removed the one build-fixes revert commit to clean up this changeset. Also removed one change that is covered in the other open PR. So basically this should be fine.

PR #68, opened 2021-04-15
Build tidy up, version from git

phlash:
Moved from #64, these are the build changes to use a sub-folder for output files, and pass through the git version/tag/branch and a dirty flag for display.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,13 +1,19 @@
 # this is licensed software, @see LICENSE file.
+
+# OpenCV & Tensorflow recommended flags for performance..
 CFLAGS = -Ofast -march=native -fno-trapping-math -fassociative-math -funsafe-math-optimizations -Wall -pthread
 LDFLAGS = -lrt -ldl
 
+# Version
+VERSION=$(shell git describe --all --always --dirty)
+CFLAGS += -D DEEPSEG_VERSION=$(VERSION)
+
 # TensorFlow
-TFBASE=tensorflow/
-TFLITE=$(TFBASE)/tensorflow/lite/tools/make/
-TFLIBS=$(TFLITE)/gen/linux_x86_64/lib/
-CFLAGS += -I $(TFBASE) -I $(TFLITE)/downloads/absl -I $(TFLITE)/downloads/flatbuffers/include -ggdb
-LDFLAGS += -L $(TFLIBS) -ltensorflow-lite -ldl
+TFBASE=tensorflow
+TFLITE=$(TFBASE)/tensorflow/lite/tools/make
+TFLIBS=$(TFLITE)/gen/linux_x86_64/lib
+TFCFLAGS += -I $(TFBASE) -I $(TFLITE)/downloads/absl -I $(TFLITE)/downloads/flatbuffers/include -ggdb
+TFLDFLAGS += -L $(TFLIBS) -ltensorflow-lite -ldl
 
 # OpenCV
 ifeq ($(shell pkg-config --exists opencv; echo $$?), 0)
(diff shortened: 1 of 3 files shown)


floe:
I think we could also merge this into main directly, no? (One minor question: what happens if it's not a git repo that the build is run in?)

phlash:
I did wonder if this was more appropriate for main branch, happy to move it across. Just tried building outside a git repo, it results in a blank string for the version (and a message on stderr during the build). Perhaps a Makefile check for an empty string and resort to something alerting?

floe:
> Perhaps a Makefile check for an empty string and resort to something alerting?

Or just default to the last release ("v0.2.0"), that's something we can update manually from time to time.

BenBE:
Moved over to experimental for now, as this does some bigger change.

Makefile
```diff
@@ -20,19 +30,34 @@ else
...
 
 $(TFLIBS)/libtensorflow-lite.a: $(TFLITE)
 	cd $(TFLITE) && ./download_dependencies.sh && ./build_lib.sh
 
 $(TFLITE):
 	git submodule update --init --recursive
 
-all: deepseg
+# Single file test progs - OpenCV deps only
+$(BIN)/%: %.cc
```
BenBE, line 56: Why not make those rules also depend on `${BIN}`?

deepseg.cc
```diff
@@ -550,7 +555,7 @@ int main(int argc, char* argv[]) {
 
 		cv::Mat test;
 		cv::cvtColor(calcinfo.raw,test,CV_YUV2BGR_YUYV);
-		cv::imshow("output.png",test);
+		cv::imshow("Deepseg:" _STR(DEEPSEG_VERSION),test);
```
BenBE, line 558:
```suggestion
		cv::imshow("DeepSeg " _STR(DEEPSEG_VERSION),test);
```

deepseg.cc
```diff
@@ -303,8 +308,8 @@ void calc_mask(calcinfo_t &info, timinginfo_t &ti) {
 
 int main(int argc, char* argv[]) {
 
-	printf("deepseg v0.2.0\n");
-	printf("(c) 2021 by floe@butterbrot.org\n");
+	printf("deepseg version:%s\n", _STR(DEEPSEG_VERSION));
```
BenBE, line 311:
```suggestion
	printf("deepseg version %s\n", _STR(DEEPSEG_VERSION));
```

Makefile
```diff
@@ -1,13 +1,23 @@
...
 CFLAGS = -Ofast -march=native -fno-trapping-math -fassociative-math -funsafe-math-optimizations -Wall -pthread
 LDFLAGS = -lrt -ldl
 
+# Version
+VERSION=$(shell git describe --all --long --always --dirty)
+ifeq ($(VERSION),)
+	VERSION=v0.2.0-no-git
+endif
+
+CFLAGS += -D DEEPSEG_VERSION=$(VERSION)
```
BenBE, line 13: This puts the version in the define without the quotes. Also prone to blanks in the tag name. Should both be quoted in the makefile and for the actual program source, so as to skip on the `__STR` macro trick.

Makefile
phlash: Good point...

deepseg.cc
phlash: Not a fan of colons, ok...done

Makefile
phlash: git does not permit tags, branches or other references with white space or escape characters in (https://git-scm.com/docs/git-check-ref-format), so I consider this safe without quoting. I often find that attempting to quote things at indeterminate levels of make/shell/compiler wrapper etc. usually results in pain, hence I try to avoid the need.

deepseg.cc
phlash: done

BenBE (commented):
I'd suggest to add all dotfiles (`.*`) as well as temporary files of different tools (`*~`, `*.bak`, `*.orig`). Other way would be to inverse and whitelist explicitly:

```
*
!*.c
!*.cc
!*.h
!Makefile
!*.md
```

Apart from this LGTM.

phlash:
done, squash merging..

Issue #67, opened 2021-04-15
Recommendation: Light wrapping


MartinKlevs:
The Meet implementation discussed [here](https://ai.googleblog.com/2020/10/background-features-in-google-meet.html) uses a method called [light wrapping](https://www.imaging-resource.com/news/2016/02/11/create-natural-looking-composite-images-using-light-wrapping-technique) to post-process the output. This removes the halo effect around the segmentation borders.

floe:
Somewhat addressed by the most recent merge from experimental, please test.

PR #72, opened 2021-04-17
Alpha-blending using smoothed mask

phlash:
Using a smoothed segmentation mask to alpha blend between background and foreground (captured video), this is a simple not-quite-light-wrapping (there is no blurred background component) fix to the hard edges we have presently. Might address #67.
For reference: https://www.premiumbeat.com/blog/what-is-light-wrapping-tips-tutorials/ provides a reasonable description of light wrapping.

### DIFF
--- a/deepseg.cc
+++ b/deepseg.cc
@@ -76,6 +76,29 @@ cv::Mat convert_rgb_to_yuyv( cv::Mat input ) {
 	return yuyv;
 }
 
+cv::Mat alpha_blend(cv::Mat srca, cv::Mat srcb, cv::Mat mask) {
+	// alpha blend two (8UC3) source images using a mask (8UC1, 255=>srca, 0=>srcb), adapted from:
+	// https://www.learnopencv.com/alpha-blending-using-opencv-cpp-python/
+	assert(srca.rows==srcb.rows);
+	assert(srca.cols==srcb.cols);
+	cv::Mat out = cv::Mat::zeros(srca.size(), srca.type());
+	uint8_t *optr = (uint8_t*)out.data;
+	uint8_t *aptr = (uint8_t*)srca.data;
+	uint8_t *bptr = (uint8_t*)srcb.data;
+	uint8_t *mptr = (uint8_t*)mask.data;
+	int npix = srca.rows * srca.cols;
+	for (int pix=0; pix<npix; ++pix) {
+		// blending weights
+		float aw=(float)(*mptr)/255.0, bw=1.0-aw;
+		// blend each channel byte
+		*optr = (uint8_t)( (float)(*aptr)*aw + (float)(*bptr)*bw ); ++aptr; ++bptr; ++optr;
+		*optr = (uint8_t)( (float)(*aptr)*aw + (float)(*bptr)*bw ); ++aptr; ++bptr; ++optr;
+		*optr = (uint8_t)( (float)(*aptr)*aw + (float)(*bptr)*bw ); ++aptr; ++bptr; ++optr;
+		++mptr;
+	}
+	return out;
+}
+
 // Tensorflow Lite helper functions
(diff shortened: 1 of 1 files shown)


deepseg.cc
```diff
@@ -76,6 +76,29 @@ cv::Mat convert_rgb_to_yuyv( cv::Mat input ) {
 	return yuyv;
 }
 
+cv::Mat alpha_blend(cv::Mat srca, cv::Mat srcb, cv::Mat mask) {
```
BenBE, line 79: Isn't there something for this already in OpenCV?

deepseg.cc
```diff
@@ -76,6 +76,29 @@ cv::Mat convert_rgb_to_yuyv( cv::Mat input ) {
...
+	uint8_t *bptr = (uint8_t*)srcb.data;
+	uint8_t *mptr = (uint8_t*)mask.data;
+	int npix = srca.rows * srca.cols;
+	for (int pix=0; pix<npix; ++pix) {
+		// blending weights
+		float aw=(float)(*mptr)/255.0, bw=1.0-aw;
+		// blend each channel byte
+		*optr = (uint8_t)( (float)(*aptr)*aw + (float)(*bptr)*bw ); ++aptr; ++bptr; ++optr;
+		*optr = (uint8_t)( (float)(*aptr)*aw + (float)(*bptr)*bw ); ++aptr; ++bptr; ++optr;
+		*optr = (uint8_t)( (float)(*aptr)*aw + (float)(*bptr)*bw ); ++aptr; ++bptr; ++optr;
```
BenBE, line 96: Put increments inline with the read/write accesses. Use some bit-shift/multiply algorithm to avoid slow floating point arithmetic.

deepseg.cc
```diff
@@ -76,6 +76,29 @@ cv::Mat convert_rgb_to_yuyv( cv::Mat input ) {
...
+		++mptr;
```
BenBE, line 97: Include with line 92 above as inline increment.

deepseg.cc
phlash: Nothing obvious: it has a constant alpha blend function `addWeighted`, and masked binary blending via `copyTo` or `add` each with a mask parameter.
There is the usual half answered SO question: https://stackoverflow.com/questions/41263010/how-to-blend-images-using-a-mask-in-opencv and an unanswered OpenCV forum one: https://answers.opencv.org/question/28439/blending-two-images-with-masked-roi/

deepseg.cc
phlash: Yeah, this could be tidier, I cut/paste... will fix.
Now that I'm not using a float mask (I am in my fork where this came from), I can use integer math too.

deepseg.cc
phlash: Yep..

deepseg.cc
BenBE: Searching their "new" forums I stumbled upon:
https://forum.opencv.org/t/opencv-java-how-to-superimpose-three-images-that-have-the-same-size/857/2

This links to this blog post:
https://learnopencv.com/alpha-blending-using-opencv-cpp-python/

phlash: Heh - literally where I got my code from..

BenBE (approved):
LGTM.

One possible option might be to avoid the division by 255 and try to get this somehow to use a division by 256. But that for another time …

phlash:
Initial timing measurements indicate this adds 0.5-1.0 msec to the post-processing compared to `bg.copyTo()` operation.

floe:
Ah, that's a good idea, thanks. What happens if I cherry-pick this into main, would that mess anything up down the line?

BenBE:
No, this should be fine. Git detects when you merge things where you've had the same patch applied already.

Issue #29, opened 2021-01-15
Segmentation result quality is subpar


nylocx:
Hi, first thanks for the project and trying to bring background removal to linux, but for me the result is really bad and I don't know why.
For me it often looks like that:
![image](https://user-images.githubusercontent.com/3329124/104724751-64bc2d00-5731-11eb-8284-c1f5b1f6a0fa.png)

Here is the debug output of the deepseg binary:
```
./deepseg -d -d -c /dev/video0 -v /dev/video4
deepseg v0.2.0
(c) 2021 by floe@butterbrot.org
https://github.com/floe/deepbacksub
debug:  2
ccam:   /dev/video0
vcam:   /dev/video4
width:  640
height: 480
back:   images/background.png
threads:2
model:  models/segm_full_v679.tflite

vid_format->type                = 2
vid_format->fmt.pix.width       = 640
vid_format->fmt.pix.height      = 480
vid_format->fmt.pix.pixelformat = 1448695129
vid_format->fmt.pix.sizeimage   = 614400
vid_format->fmt.pix.field       = 1
vid_format->fmt.pix.bytesperline= 1280
vid_format->fmt.pix.colorspace  = 8

tensor #0: 1
tensor #0: 144
tensor #0: 256
tensor #0: 3
tensor #244: 1
tensor #244: 144
tensor #244: 256
tensor #244: 2
```

And this is from v4l2-ctl --all -d /dev/video0
```
Driver name      : uvcvideo
Video input : 0 (Camera 1: ok)
Format Video Capture:
        Width/Height      : 640/480
        Pixel Format      : 'YUYV' (YUYV 4:2:2)
        Field             : None
        Bytes per Line    : 1280
        Size Image        : 614400
        Colorspace        : sRGB
        Transfer Function : Rec. 709
        YCbCr/HSV Encoding: ITU-R 601
        Quantization      : Default (maps to Limited Range)
        Flags             : 
```
```
opencv_version
4.5.1
```
So looks ok for me

Any ideas where I should start debugging this?

floe:
Two things to try: 

- improve overall lighting for the scene (I've had similar issues with the Google Meet network)
- try another network from models/ (e.g. `-m models/deeplabv3_257_mv_gpu.tflite` - will be slower, though)

shannonlowder:
[...]

PR #77, opened 2021-05-08
Normalize correctly for segm_ / selfie

vekkt0r:
Expected input is [0.0, 1.0] according to MediaPipe / MLKit model cards. Performance with negative inputs is really bad with segm_ model.

Have not checked any of the other models, maybe should add an conditional around this?

### DIFF
--- a/deepseg.cc
+++ b/deepseg.cc
@@ -244,8 +244,8 @@ void calc_mask(calcinfo_t &info, timinginfo_t &ti) {
 		in_u8_rgb = filtered;
 	}
 
-	// convert to float and normalize values to [-1;1]
-	in_u8_rgb.convertTo(info.input,CV_32FC3,1.0/128.0,-1.0);
+	// convert to float and normalize values to [0;1]
+	in_u8_rgb.convertTo(info.input,CV_32FC3,1.0/255.0);
 	ti.openns=timestamp();
 
 
--- a/deepseg.cc
+++ b/deepseg.cc
@@ -148,8 +148,23 @@ typedef struct {
 	pthread_mutex_t lock;
 } capinfo_t;
 
+enum class modeltype_t {
+	Unknown,
+	BodyPix,
+	DeepLab,
+	GoogleMeetSegmentation,
+	MLKitSelfie,
+};
+
+struct normalization_t {
+	float scaling;
+	float offset;
+};
+
 typedef struct {
 	const char *modelname;
+	modeltype_t modeltype;
+	normalization_t norm;
 	size_t threads;
 	size_t width;
 	size_t height;
@@ -188,6 +203,36 @@ void *grab_thread(void *arg) {
 		pthread_mutex_unlock(&ci->lock);
 	}
 	return NULL;
+}
+
+modeltype_t get_modeltype(const char* modelname) {
+	if (strstr(modelname, "body-pix")) {
+		return modeltype_t::BodyPix;
+	}
+	else if (strstr(modelname, "deeplab")) {
+		return modeltype_t::DeepLab;
+	}
+	else if (strstr(modelname, "segm_")) {
+		return modeltype_t::GoogleMeetSegmentation;
+	}
+	else if (strstr(modelname, "selfie")) {
+		return modeltype_t::MLKitSelfie;
+	}
+	return modeltype_t::Unknown;
+}
+
+normalization_t get_normalization(modeltype_t type) {
+	// TODO: This should be read out from actual mode metadata instead
+	switch (type) {
+		case modeltype_t::DeepLab:
+			return normalization_t{.scaling = 1/127.5, .offset = -1};
+		case modeltype_t::BodyPix:
+		case modeltype_t::GoogleMeetSegmentation:
+		case modeltype_t::MLKitSelfie:
+		case modeltype_t::Unknown:
+		default:
+			return normalization_t{.scaling = 1/255.0, .offset = 0};
+	}
 }
 
 void init_tensorflow(calcinfo_t &info) {
@@ -244,10 +289,9 @@ void calc_mask(calcinfo_t &info, timinginfo_t &ti) {
 		in_u8_rgb = filtered;
 	}
 
-	// convert to float and normalize values to [0;1]
-	in_u8_rgb.convertTo(info.input,CV_32FC3,1.0/255.0);
+	// convert to float and normalize to values expected by model
+	in_u8_rgb.convertTo(info.input,CV_32FC3,info.norm.scaling,info.norm.offset);
 	ti.openns=timestamp();
-
 
 	// Run inference
 	TFLITE_MINIMAL_CHECK(interpreter->Invoke() == kTfLiteOk);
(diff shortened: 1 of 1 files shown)


phlash:
Tried this locally, no obvious difference on my camera in daylight, for any of the models, will wait for others to try as well, and retest in lower light conditions..

We might wish to add the model cards for segm_ and deeplab models here though as reference?

phlash:
Looks like deeplabv3 doesn't have a model card (yet), but there is documentation here: https://tfhub.dev/tensorflow/lite-model/deeplabv3/1/metadata/2
Google Meet Segmentation model (Xeno) has a card: https://mediapipe.page.link/meet-mc

BenBE:
If these ranges differ for those models, why not put some small configuration together that accompanies each model file and provides these settings? This configuration could also contain settings for common post-processing steps like blur radii, additional filters or input dimensions. That way we'd have to maintain less per-network code and also keep the settings near each network they apply to. TBH, I kinda hate those various lines `if(strncmp(filename, "segm_", 5)) …` that already accumulated in the source.

phlash:
I wonder if the built-in metadata would work well enough for model processing: https://www.tensorflow.org/lite/convert/metadata ?
This would avoid any issues with separate files getting out of sync / renamed etc.

Other application config (such as blur radius) would be shared and makes sense to live by itself IMO (and could be exposed as part of a library API - a direction I was headed for with my [tflite_extract fork](https://github.com/phlash/backscrub/tree/tflite-extract), and should probably be a separate PR ;)

vekkt0r:
> Tried this locally, no obvious difference on my camera in daylight, for any of the models, will wait for others to try as well, and retest in lower light conditions..
> 
Thanks for testing. Just the Google Meet network that showed problems for me, first few frames was ok before backlight control kicked in and made the image a bit darker. Sounds similar to problems discussed in #29

vekkt0r:
> If these ranges differ for those models, why not put some small configuration together that accompanies each model file and provides these settings? This configuration could also contain settings for common post-processing steps like blur radii, additional filters or input dimensions. That way we'd have to maintain less per-network code and also keep the settings near each network they apply to. TBH, I kinda hate those various lines `if(strncmp(filename, "segm_", 5)) …` that already accumulated in the source.

Hehe fully agree on those strstr checks. Need to hold back my urge to "optimise" every time I glance the code. But now I might actually have reason to remove them 😉

> I wonder if the built-in metadata would work well enough for model processing: https://www.tensorflow.org/lite/convert/metadata ?
> This would avoid any issues with separate files getting out of sync / renamed etc.

Sounds like a nice solution, but is the metadata normally included in models? Sounds like models with metadata should essentially be a zip file?

phlash:
> Sounds like a nice solution, but is the metadata normally included in models? Sounds like models with metadata should essentially be a zip file?

Not in the models we're currently using, but only because we didn't need it. models with metadata in are available from the Google model zoo (https://tfhub.dev/).

I'm currently looking at sane ways to read the metadata without adding a new dependency and build pain (it needs Bazel) for the `tf_lite_support` library that's officially required. So far, getting a blob of metadata out works ok (see snippet below), since TfLite already supports random blobs in a file, however that blob then needs parsing (from `flatbuffers` format, schema here: https://github.com/tensorflow/tflite-support/blob/master/tensorflow_lite_support/metadata/metadata_schema.fbs) to pull out the input normalization constants.. Currently poking through this: https://google.github.io/flatbuffers/md__internals.html and a hex dump of the raw buffer.
```c++
int init_tensorflow(...) {
	//...
	auto model = flatmodel->GetModel();
	auto *md = model->metadata();
	if (md) {
		for (uint32_t mid=0; mid < md->size(); ++mid) {
			const auto meta = md->Get(mid);
			printf("found: %s\n", meta->name()->c_str());
			if (meta->name()->str() != "TFLITE_METADATA")
				continue;
			// grab raw buffer and dump it..
			const flatbuffers::Vector<uint8_t> *pvec = model->buffers()->Get(meta->buffer())->data();
			printf("metadata dump (size=0x%X)\n", pvec->size());
			parse_metadata(pvec->data(), pvec->size());  // currently just a hex dump :)
		}
	}
```
This might turn out to be horribly fragile though!

Given your original intention of simply fixing the normalization, perhaps grit your teeth and use the `strcmp(...)` for now, while we think about Bazel and additional (beta) libraries!

BenBE:
@phlash Could you add an issue for the Bazel parsing stuff? Just so we can concentrate on the normalization issue here and have the "Make things shiny" in another ticket?

floe:
I was originally planning to create an abstract `Model` class and derive the individual model wrappers from that, but never got around to that. Would that be an option to deal with this?

BenBE:
Sounds like a reasonable plan. Although you might want to have some central code selecting the right class to instantiate. But that's doable.

deepseg.cc
```diff
@@ -113,6 +113,18 @@ const std::vector<std::string> labels = { "background", "aeroplane", "bicycle",
 const size_t cnum = labels.size();
 const size_t pers = std::distance(labels.begin(), std::find(labels.begin(),labels.end(),"person"));
 
+enum class ModelType {
+	BodyPix,
+	DeepLab,
+	GoogleMeetSegmentation,
+	MLKitSelfie,
+};
```
BenBE, line 121: More consistent naming? Also following STL naming conventions snake_case is preferred.
```suggestion
enum class model_type {
	BodyPix,
	DeepLab,
	GoogleMeetSegmentation,
	MLKitSelfie,
};
```

deepseg.cc
```diff
@@ -113,6 +113,18 @@ const std::vector<std::string> labels = { "background", "aeroplane", "bicycle",
...
+
+typedef struct {
+	float mean;
+	float std;
+} meta_t;
```
BenBE, line 126:
```suggestion
struct stat_data {
	float mean;
	float std;
};
```

deepseg.cc
```diff
@@ -190,6 +204,33 @@ void *grab_thread(void *arg) {
...
+		return ModelType::DeepLab;
+	}
+	else if (strstr(modelname, "segm_")) {
+		return ModelType::GoogleMeetSegmentation;
+	}
+	else if (strstr(modelname, "selfie")) {
+		return ModelType::MLKitSelfie;
+	}
+	fprintf(stderr, "Unknown model type '%s'.\n", modelname);
+	exit(1);
```
BenBE, line 221: Preferably have an `unknown` model type normalized `0..1` and don't exit wildly from within functions. Normally a caller wouldn't expect a function call like `getSomething` to terminate your program.

deepseg.cc
```diff
@@ -190,6 +204,33 @@ void *grab_thread(void *arg) {
...
+}
+
+meta_t get_meta(ModelType type) {
+	// TODO: This should be read out from actual mode metadata instead
+	if (type == ModelType::DeepLab) {
+		return meta_t{.std = 127.5, .mean = 127.5};
```
BenBE, line 227: Wouldn't be the better description `scaling` and `offset` for those two fields?

`mean` and `std` (st(d)dev) usually have a completely different meaning. Or if those names come from the model files, this should be documented as part of the type declaration for `meta_t`.

deepseg.cc
```diff
@@ -257,7 +297,7 @@ void calc_mask(calcinfo_t &info, timinginfo_t &ti) {
 	uint8_t* out = (uint8_t*)info.ofinal.data;
 
 	// find class with maximum probability
-	if (strstr(info.modelname,"deeplab")) {
+	if (info.modeltype == ModelType::DeepLab) {
```
BenBE, line 300: What about using `switch`/`case` here? That way the compiler could actually check for completeness by testing if all cases are handled.

BenBE (changes requested):

deepseg.cc
vekkt0r: Yeah I was trying to keep them as in the metadata, but would make more sense with your suggestions. Will change!

deepseg.cc
vekkt0r: yes!

deepseg.cc
vekkt0r: Should the enum values actually be lower case also if following the STL conventions?

deepseg.cc
vekkt0r: Hm stat as in statistics? I would prefer metadata.

Agree on removing the C-style typedefs, included for consistency with rest of code.

phlash: Apologies for the `typedef` things - that was me with a C head on. I'll pop in a chore task to tidy up syntax for C++11 stds.

deepseg.cc
phlash: Actually as we use them (scale, offset) with OpenCV they are different (although derived) from the raw metadata, so different names make sense to me.

deepseg.cc
phlash: Also avoids the '=' when '==' was intended in a boolean expression bug :wink:

deepseg.cc
BenBE: Good question. Strictly speaking yes, but given the current names mesh quite well from a descriptiveness PoV converting them to snake-case would probably lose some of the appeal.

deepseg.cc
BenBE: See also my other note regarding this type. Yes, given only those names I first thought of `statistics`, but given the actual use further down a different name will likely be better. Just calling it metadata won't work though, as metadata usually refers to naming and other properties. For this the current structure doesn't look "convincing" enough to warrant that name.

deepseg.cc
BenBE: As an aside: To avoid the `=` vs. `==` confusion you sometimes see the constant expression being written first. That way accidentally writing just one `=` gives you a compile error (e.g. `2 == foo` vs. `2 = foo`).

deepseg.cc
vekkt0r: Ok lets keep it as is then 🙂

deepseg.cc
vekkt0r: Good point, fixed in last revision.

deepseg.cc
vekkt0r: Maybe `normalization_t` as in last rev?

phlash:
Are we happy to merge this one? No outstanding comments, LGTM.

deepseg.cc
```diff
@@ -148,8 +148,23 @@ typedef struct {
 	pthread_mutex_t lock;
 } capinfo_t;
 
+enum class modeltype {
```
BenBE, line 151: name consistency -> `modeltype_t` ?

deepseg.cc
```diff
@@ -257,7 +301,7 @@ void calc_mask(calcinfo_t &info, timinginfo_t &ti) {
 	uint8_t* out = (uint8_t*)info.ofinal.data;
 
 	// find class with maximum probability
-	if (strstr(info.modelname,"deeplab")) {
+	if (modeltype::DeepLab == info.modeltype) {
```
BenBE, line 304: Probably better of as a switch/case, thus ensuring to cover all cases. Currently missing the `modeltype::Unknown`/`default` case.

BenBE (changes requested):

deepseg.cc
vekkt0r: Wanted to keep it clean without _t since type name is used as prefix on each use. I'll add _t suffix.

vekkt0r:
> First two are a result of the type and structure field sharing the same name, changing the field to eg:`mtype` fixes these.
> Third is due to missing `default` case in switch.

Oups, sorry about that. Clang thinks different about the missing default (actually emits warning _with_ default if all values are covered if using -Weverytning).

deepseg.cc
vekkt0r: Changed to switch / case instead. 

Quit early if model is unknown since then we do not know how to parse the output.

phlash:
> > First two are a result of the type and structure field sharing the same name, changing the field to eg:`mtype` fixes these.
> > Third is due to missing `default` case in switch.
> 
> Oups, sorry about that. Clang thinks different about the missing default (actually emits warning _with_ default if all values are covered if using -Weverytning).

sorted - thanks, sometimes GCC & Clang have differing opinions :smile:

deepseg.cc
```diff
@@ -190,6 +205,36 @@ void *grab_thread(void *arg) {
...
+		return modeltype_t::MLKitSelfie;
+	}
+	return modeltype_t::Unknown;
+}
+
+normalization_t get_normalization(modeltype_t type) {
+	// TODO: This should be read out from actual mode metadata instead
+	switch (type) {
+		case modeltype_t::DeepLab:
+			return normalization_t{.scaling = 1/127.5, .offset = -1};
```
BenBE, line 228: Giving the `normalization_t` before the braces is even optional starting with C++14.

deepseg.cc
```diff
@@ -244,59 +289,61 @@ void calc_mask(calcinfo_t &info, timinginfo_t &ti) {
...
+				float exp0 = expf(tmp[2*n  ]);
+				float exp1 = expf(tmp[2*n+1]);
+				float p0 = exp0 / (exp0+exp1);
+				float p1 = exp1 / (exp0+exp1);
+				uint8_t val = (p0 < p1 ? 0 : 255);
+				out[n] = (val & 0xE0) | (out[n] >> 3);
+			}
+			break;
+		case modeltype_t::Unknown:
+			fprintf(stderr, "Unknown model type\n");
```
BenBE, line 345: What about assuming direct mask as output in this case?

BenBE (approved):

deepseg.cc
vekkt0r: Neat!

deepseg.cc
vekkt0r: I.e. use same as BodyPix?

phlash: @vekkt0r did you want to make this change before I merge the PR? I believe this is the last remaining point.

Issue #79, opened 2021-05-12
GPU Acceleration


SamSpiri:
Is it possible to use NVIDIA CUDA acceleration?
In order to use body-pix or deeplab with 15+ FPS performance.

phlash:
From the documentation here: https://www.tensorflow.org/lite/performance/delegates I don't think there is a Linux/x86 targetted GPU delegate that could be used, but happy to be proven wrong by someone!

Sticking to CPU, we may be able to improve by explicitly building the XNNPACK backend: https://blog.tensorflow.org/2020/07/accelerating-tensorflow-lite-xnnpack-integration.html

floe:
Oho, neat! The main reason for picking TFLite in the first place was that it has a workable CPU backend, "regular" Tensorflow basically requires a Nvidia GPU (at least as of ~ 1 year ago). And I guess part of the appeal of backscrub is that it doesn't require anything fancy in terms of hardware.

phlash:
I have just discovered that there is an OpenCL delegate (https://github.com/tensorflow/tensorflow/tree/v2.4.1/tensorflow/lite/delegates/gpu), which can be built via CMake (only tested by Google on Android cross-builds but _should_ work on Linux), however it's fighting me at the moment courtesy of untested Linux headers, and _yet-another-flatbuffers-schema_ that needs compiling..

SamSpiri:
XNNPACK results you've got are very promising!

I was looking through internet and found that tflite models are not so good for GPU acceleration.
It is much better in TF no the LITE version. But since the whole project is based on tflite it probably will take a lot of effort to refactor to TF. Or am I wrong?

SamSpiri:
btw: https://github.com/mikan3rd/react-body-pix-sample

phlash:
> I was looking through internet and found that tflite models are not so good for GPU acceleration.

Thanks for the feedback and research @SamSpiri. Given tflite is built for use on mobile devices and specifically for their GPUs I'm a little surprised you found tflite models are 'not so good for GPU' (perhaps it's that desktop GPUs might have support for 64bit floats, mobile tends not to?) - got any links to what you discovered please?

As @floe notes above, this project uses tflite so it can operate _without_ a GPU, and switching to full-fat Tensorflow will kill that off, so for now I think we're likely to stay with tflite. Code-wise, it wouldn't be particularly difficult/different to use full TF, and as/when we get a more pluggable structure I would expect someone to write the appropriate plugin :smile:

[...]

PR #81, opened 2021-05-17
GPU and DOCKER

SamSpiri:
That's @phlash 's gpu integration https://github.com/phlash/backscrub/tree/xnnpack-test
discussed in https://github.com/floe/backscrub/issues/79

Add Dockerfile and docker-compose.yml in order to quickly have everything ready to compile the project.
To try it out: Edit docker-compose.yml to set correct arguments and run `docker-compose up`

Add some more command line arguments
-G to enable GPU
-s to set FPS
-T to set threshold_probability, which was with FIXME comment.
But I didn't notice any change while going from 0.2 to 0.8. Bug?

Accidentally commited the starwars room background.


BenBE:
Having a Docker file doesn't relief you from documenting step-by-step how to setup a proper environment.

Also the GPU options should only be available with builds that enable GPU support. Having such flags available may be part of some compile option list, or may be shown as `GPU: unavailable` instead of the usual `GPU: yes`/`GPU: no`

phlash:
@SamSpiri thanks for the enthusiasm, glad the GPU hackery worked at all! You might want to break this PR down a bit, there's a quite a lot going on:
 * Shift to CMake builds - we have a PR for that (#50), not sure where that is with avoiding dependency builds on every call, I simply started there and worked out how to turn on other features in a TfLite build.
 * Docker wrapper - might be better as a separate git repo that references backscrub as a sub-project. I'm not sure who would look after Docker stuff in this tree :question: so for now you are likely to retain ownership of this for yourself or others who want a Dockerised build.
   * Docker base image - you've picked Cuda/Nvidia, which is fine if that's what you've got, there will be AMD and possibly others..
 * backscrub code changes - as discussed in #79, and echoing @BenBE comments above, we haven't made any decision to include GPU acceleration in the `main` branch or to build with CMake yet. It's now proven to be possible, but we ought to consider user experience (eg: should there be a graceful fallback if GPU is not usable - right now it crashes out?), and any XNNPACK or GPU support relies on a CMake build which is not officially in yet (although I think this will be 'real soon now' :tm: )

Please be patient with us, as we've yet to arrange an IRL conversation to talk this stuff through :smile:

SamSpiri:
That PR is just a quick share of my progress so far.
I didn't mean to merge in with all that stuff. That's too much for one PR indeed.
I'd wait for the CMake merged in, and than we could base on it for further improvements like XNNPACK and GPU.

I also added background blurring to the mix (only locally yet). Got it from another pull request. And added option to use GPU for blurring. It actually opens the pandora box to do more stuff on gpu. Looking forward - also a mask blending on gpu would be cool.

phlash:
@SamSpiri can this PR be closed? Now we have CMake builds, you should be in a position to submit various individual changes as suggested above...

phlash:
closing due to inactivity

PR #83, opened 2021-05-19
Enable XNNPACK in CMake builds

phlash:
enables use of XNNPACK delegate by default on CPU, which provides ~2x performance improvement.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,6 +3,10 @@ project(deepbacksub CXX)
 
 find_package(OpenCV REQUIRED COMPONENTS core imgproc imgcodecs video videoio highgui)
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -1,8 +1,12 @@
 cmake_minimum_required(VERSION 3.16)
-project(deepbacksub CXX)
+project(backscrub CXX)
```
BenBE, line 2: The name change should be its own separate commit.

BenBE (approved):
Code-wise LGTM. Didn't run/test locally yet.

CMakeLists.txt
phlash: Good point, I'll put that elsewhere..

Issue #84, opened 2021-05-21
Upgrading to Tensorflow Lite 2.5


vongillus:
Right now the backscrub code relies on TF Lite 2.4.1.

When I tested the new code of @Phlash with XNNPACK on ARM architecture (aarch64), there is a bug in the version of XNNPACK packaged with TF 2.4.1 which prevents compiling it:

https://github.com/google/XNNPACK/issues/987

On the latest version of TF Lite 2.5, I could compile backscrub fork with XNNPACK on ARM without issues.

phlash:
I guess we need a few people to give this a go and report back if it's ok?

@vongillus are you happy to put a PR in for the version change, then testers can pull that PR down to test, I use these incantations:
```bash
%> git checkout main
%> git fetch origin pull/<PR number>/head:<local branch name>
%> git checkout <local branch name>
%> cmake --build build/
%> build/deepseg -d -d -c <your camera device> -v <your loopback device>
```

phlash:
Looks good in #61, I'll also test against `main` and put in the PR if it works for me.

PR #86, opened 2021-05-24
Libbackscrub - separation into library and wrapper application

phlash:
This PR moves all the functionality that generates a mask from an input image into a separate library (`libbackscrub`), leaving the I/O and debug functions in the wrapper application (`deepseg`). It builds with both `make` and `cmake`, the latter also providing faster CPU operation via XNNPACK and an export descriptor (`BackscrubTargets.cmake`) for use in other `cmake` projects that reference the library (such as my sample `obs-backscrub` plugin).

There are a number of `TODO` markers in the code where I think more attention is required, in particular around the shared data structure `calcinfo_t` that is both poorly named and fragile with respect to storage allocation. It may be better to hide more of the detail within the library (some already is via `backscrub_ctx_t` internal type), even removing the shared structure in favour of more parameters to the `init` function - __this is now done__

NB: I have chosen not to alter the separated function names yet, although we might like to prefix them with eg: `lbs_` or put them in a separate namespace - __this is now done__


CMakeLists.txt
```diff
@@ -20,29 +23,81 @@ add_compile_definitions(TFLITE_BUILD_WITH_XNNPACK_DELEGATE)
 
 # This assumes that tensorflow got checked out as submodule.
 # True, if `git clone` had the additional option `--recursive`.
-add_subdirectory(tensorflow/tensorflow/lite 
+add_subdirectory(tensorflow/tensorflow/lite
```
BenBE, line 26: Could we make this path dynamic, so that people can override this with a parameter to `cmake`?
This also allows to bail out if TFlite is not properly checked out/cannot be found.

CMakeLists.txt
```diff
@@ -20,29 +23,81 @@ add_compile_definitions(TFLITE_BUILD_WITH_XNNPACK_DELEGATE)
...
   "${CMAKE_CURRENT_BINARY_DIR}/tensorflow-lite" EXCLUDE_FROM_ALL)
 
 add_compile_definitions(DEEPSEG_VERSION=${DEEPSEG_VERSION})
 
+add_library(backscrub
+  libbackscrub.cc
+  transpose_conv_bias.cc)
```
BenBE, line 33: Maybe split out the lib sources into a `lib/` or `src/lib/` sub directory? Thus moving all source files out from the main source directory.

CMakeLists.txt
```diff
@@ -20,29 +23,81 @@ add_compile_definitions(TFLITE_BUILD_WITH_XNNPACK_DELEGATE)
...
+target_link_libraries(backscrub
+  tensorflow-lite ${CMAKE_DL_LIBS}
+  opencv_core
+  opencv_imgproc
+)
+
 add_executable(deepseg
   deepseg.cc
-  loopback.cc 
+  loopback.cc
```
BenBE, line 43: Maybe also put the handling of the loopback device into the library. Can be a separate set of API calls from the user side, but given the minimal interface required (`bs_vcam_open`, `bs_vcam_openfd`, `bs_vcam_frame_push`, `bs_vcam_close`, plus maybe some helper functions) this might be worth the split. Or have the vcam stuff split into its own library too.

CMakeLists.txt
```diff
@@ -20,29 +23,81 @@ add_compile_definitions(TFLITE_BUILD_WITH_XNNPACK_DELEGATE)
...
+export(TARGETS
+  backscrub
+  ${BACKSCRUB_DEPS}
+  FILE BackscrubTargets.cmake)
 
 install(TARGETS deepseg)
-install(DIRECTORY models 
-    DESTINATION ${CMAKE_INSTALL_PREFIX}/share/deepbacksub 
+install(TARGETS backscrub)
+install(FILES libbackscrub.h DESTINATION include)
```
BenBE, line 100: Maybe use a subdirectory like `/usr/include/backscrub` to avoid polluting the main `/usr/include` directory with our files.

libbackscrub.cc
```diff
@@ -0,0 +1,272 @@
...
+	va_list ap;
+	va_start(ap, fmt);
+	if (info.ondebug)
+		info.ondebug(info.caller_ctx, fmt, ap);
+	else
+		vfprintf(stderr, fmt, ap);
+	va_end(ap);
+}
+
+using namespace tflite;
```
BenBE, line 26: Absolutely not a fan. Especially on a global source level.

libbackscrub.cc
```diff
@@ -0,0 +1,272 @@
+/* This is licenced software, @see LICENSE file.
```
BenBE, line 1: Further split this file into the TFlite parts and the actual mask gen parts.

libbackscrub.h
```diff
@@ -0,0 +1,54 @@
...
+// tested against tensorflow lite v2.4.1 (static library)
+#ifndef _LIBBACKSCRUB_H
+#define _LIBBACKSCRUB_H
+
+// for cv::Mat and related types
+#include <opencv2/core/core.hpp>
+#include <opencv2/imgproc/imgproc.hpp>
+#include <stdarg.h>
+
+#define TFLITE_MINIMAL_CHECK(x)                              \
```
BenBE, line 14: There should be no need to export this symbol to users of our library. If there is, the abstraction is broken.

libbackscrub.h
```diff
@@ -0,0 +1,54 @@
...
+	cv::Mat mask;
+	cv::Mat mroi;
+	cv::Mat raw;
+	cv::Mat ofinal;
+	cv::Size blur;
+	float ratio;
+	void *backscrub_ctx;	// opaque context used by libbackscrub
+} calcinfo_t;
+
+extern int init_tensorflow(calcinfo_t &info);
```
BenBE, line 50: `bs_maskgen_init`

libbackscrub.h
```diff
@@ -0,0 +1,54 @@
...
+extern void drop_tensorflow(calcinfo_t &info);
```
BenBE, line 51: `bs_maskgen_free`

libbackscrub.h
```diff
@@ -0,0 +1,54 @@
...
+extern int calc_mask(calcinfo_t &info);
```
BenBE, line 52: `bs_maskgen_process`???

BenBE (changes requested):
`calcinfo_t` should be more opaque to the library user. Instead the library should provide proper getters/setters for properties to update internal structures upon a change to settings.

CMakeLists.txt
phlash: Sure - I'll use an overridable variable - might want this back in main too?

CMakeLists.txt
phlash: Sure - `lib/` and `app/` ok?

CMakeLists.txt
phlash: I'll go with a separate library, given the initial use case of incorporating backscrub into OBS Studio which will do all the the I/O, and provides a cut point to port to other operating systems that have other video loopback APIs.

CMakeLists.txt
phlash: Probably a safer option, this comes from existing CMake (and I've never actually done an install :smile: )

libbackscrub.cc
phlash: Yeah, I don't like `using` either - nuking.

libbackscrub.cc
phlash: Yep - one source file => one obvious job.

libbackscrub.h
phlash: Agreed - another hangover from original code, please bring your own assertion checker outside the library :smile:

libbackscrub.h
phlash: Agreed - `bs_maskgen_` prefix for this library, `bs_videoio_` for loopback (or other equivalent)

libbackscrub.h
phlash: Might prefer `new/delete` semantics for this to preserve the balance?

libbackscrub.h
phlash: ACK.

Makefile
```diff
@@ -11,11 +11,12 @@ ifeq ($(VERSION),)
 	VERSION=v0.2.0-no-git
 endif
 
-CFLAGS += -D DEEPSEG_VERSION=$(VERSION)
+CFLAGS += -D DEEPSEG_VERSION=$(VERSION) -I .
```
BenBE, line 14:
```suggestion
CFLAGS += -DDEEPSEG_VERSION=$(VERSION) -I.
```

Makefile
```diff
@@ -53,13 +54,28 @@ $(BIN)/libbackscrub.a: $(BIN)/libbackscrub.o $(BIN)/transpose_conv_bias.o $(TFLI
...
+$(BIN)/%.o: app/%.cc $(TFDOWN)
+	g++ $< ${CFLAGS} ${TFCFLAGS} -c -o $@
 
 # As cloned, TFLite needs building into a static library, as per:
 # https://github.com/tensorflow/tensorflow/tree/master/tensorflow/lite/tools/make
+# we split this into download (pre-compile) and build (pre-link)
+$(TFDOWN): $(TFLITE)
+	cd $(TFLITE) && ./download_dependencies.sh
+
 $(TFLIBS)/libtensorflow-lite.a: $(TFLITE)
```
BenBE, line 77: Shouldn't this depend on `${TFDOWN}` (or better a marker file to avoid making a directory the target and thus this rule implicitly `.PHONY`).

CMakeLists.txt
```diff
@@ -1,6 +1,10 @@
 cmake_minimum_required(VERSION 3.16)
```
BenBE, line 1: Do we (or any of our dependencies) use any features from 3.16, or could we potentially mark this compatible with a older release?

CMakeLists.txt
```diff
@@ -1,6 +1,10 @@
...
 project(backscrub CXX)
 
+if(NOT DEFINED TENSORFLOW)
+  set(TENSORFLOW tensorflow)
```
BenBE, line 5: Consistent naming with `Makefile`? --> `TFBASE`?

BenBE (changes requested):

CMakeLists.txt
phlash: Tensorflow requires this

CMakeLists.txt
phlash: Can do - might go the other way?

Makefile
phlash: I needed a folder (with a datestamp) to trigger the dowload, which makes rebuilds work as expected (no download when already done - don't think directories are implicitly .PHONY?). The dependency back to `$(TFLITE)` is to trigger the `git submodue update`. Seems to all work...

Makefile
phlash: Heh - sure.

CMakeLists.txt
BenBE: Just important to try to have both in sync. Direction doesn't matter.

CMakeLists.txt
BenBE: Okay. Just wanted to make sure this is set for a reason and not just arbitrarily.

Makefile
BenBE: Clear from the intention. Just not sure if the `$(TFLIBS)/libtensorflow-lite.a` should better depend on `${TFDOWN}` instead to force completion of the download step before `$(TFLIBS)/libtensorflow-lite.a` is tried to be built. This is likely to work with single threaded builds, but asking to break with parallel builds.

libbackscrub.h
BenBE: Indifferent about that. Only thing I ask for is for the name scheme to be somewhat consistent. ;-)

Makefile
```diff
@@ -43,12 +44,38 @@ clean:
...
-$(BIN)/deepseg: $(TFLIBS)/libtensorflow-lite.a deepseg.cc loopback.cc transpose_conv_bias.cc
+# Primary binaries - special deps
+$(BIN)/deepseg: app/deepseg.cc $(BIN)/libbackscrub.a $(BIN)/libvideoio.a
 	g++ $^ ${CFLAGS} ${TFCFLAGS} ${LDFLAGS} ${TFLDFLAGS} -o $@
 
+# Unusual archive munging here is the nicest way to ensure we have Tensorflow Lite & our library code
+# easily accessible through one static library
+$(BIN)/libbackscrub.a: $(BIN)/libbackscrub.o $(BIN)/transpose_conv_bias.o $(TFLIBS)/libtensorflow-lite.a
+	cp -p $(TFLIBS)/libtensorflow-lite.a $@
+	ar rv $@ $(BIN)/libbackscrub.o $(BIN)/transpose_conv_bias.o
```
BenBE, line 55: Why not leave them separate and just pass both of them to the linker later on? If we pass `libbackscrub.a` followed by its dependencies, we can avoid this hack. As an aside: If we want to skip managing libs ourselves we could ask `libtool` to do that job for us.

lib/libbackscrub.cc
```diff
@@ -0,0 +1,270 @@
+/* This is licenced software, @see LICENSE file.
+ * Authors - @see AUTHORS file.
+==============================================================================*/
+
+// tested against tensorflow lite v2.4.1 (static library)
```
BenBE, line 5: Prone to be outdated. Especially in light of #84 …

lib/libbackscrub.cc
```diff
@@ -0,0 +1,270 @@
...
+};
+
+static cv::Mat getTensorMat(calcinfo_t &info, int tnum) {
+
+	backscrub_ctx_t &ctx = *((backscrub_ctx_t *)info.backscrub_ctx);
+	TfLiteType t_type = ctx.interpreter->tensor(tnum)->type;
+	TFLITE_MINIMAL_CHECK(t_type == kTfLiteFloat32);
+
+	TfLiteIntArray* dims = ctx.interpreter->tensor(tnum)->dims;
+	if (info.debug)
```
BenBE, line 54: This check should be in `_dbg` …

lib/libbackscrub.h
```diff
@@ -0,0 +1,54 @@
...
+// for cv::Mat and related types
+#include <opencv2/core/core.hpp>
+#include <opencv2/imgproc/imgproc.hpp>
+#include <stdarg.h>
+
+#define TFLITE_MINIMAL_CHECK(x)                              \
+  if (!(x)) {                                                \
+	fprintf(stderr, "Error at %s:%d\n", __FILE__, __LINE__); \
+	exit(1);                                                 \
+  }
```
BenBE, line 18: Consistent indentation

lib/libbackscrub.h
```diff
@@ -0,0 +1,54 @@
...
+	// Used by libbackscrub / callbacks (eg: adjusting blur size)
+	// XXX:TODO: probably too much coupling - PAA
+	cv::Mat input;
+	cv::Mat output;
+	cv::Rect roidim;
+	cv::Mat mask;
+	cv::Mat mroi;
+	cv::Mat raw;
+	cv::Mat ofinal;
+	cv::Size blur;
```
BenBE, line 45: Possibly avoid C++ here, to allow the header to be a plain C interface.
Also should include an `extern "C" {}` block around everything in here. Otherwise function names of the library will be mangled and thus interop with other languages will be a hassle.

lib/libbackscrub.h
```diff
@@ -0,0 +1,54 @@
...
+ * Authors - @see AUTHORS file.
+==============================================================================*/
+
+// tested against tensorflow lite v2.4.1 (static library)
+#ifndef _LIBBACKSCRUB_H
+#define _LIBBACKSCRUB_H
+
+// for cv::Mat and related types
+#include <opencv2/core/core.hpp>
+#include <opencv2/imgproc/imgproc.hpp>
```
BenBE, line 11: Maybe avoid those here in the public interface to allow for a plain C interface.

lib/libbackscrub.h
```diff
@@ -0,0 +1,54 @@
...
+#include <stdarg.h>
```
BenBE, line 12: I prefer to have system headers go before other dependencies and before headers from the own project. In this case:
* `libc`: System libs
* `opencv`, `tflite`: external dependencies
* `libvideoio`: internal dependencies
* `libbackscrub`: own library

lib/libbackscrub.cc
```diff
@@ -0,0 +1,270 @@
...
+	backscrub_ctx_t &ctx = *((backscrub_ctx_t *)info.backscrub_ctx);
+	// Load model
+	ctx.model = tflite::FlatBufferModel::BuildFromFile(info.modelname);
+	TFLITE_MINIMAL_CHECK(ctx.model != nullptr);
+	// Determine model type and normalization values
+	ctx.modeltype = get_modeltype(info.modelname);
+	ctx.norm = get_normalization(ctx.modeltype);
+	if (modeltype_t::Unknown == ctx.modeltype) {
+		_dbg(info, "Unknown model type '%s'.\n", info.modelname);
+		return 0;
```
BenBE, line 123: Potentially leaks `ctx` instance. On failure `info` should hold no allocated resources.

lib/libbackscrub.cc
```diff
@@ -0,0 +1,270 @@
...
+	}
+	// Build the interpreter
+	tflite::ops::builtin::BuiltinOpResolver resolver;
+	// custom op for Google Meet network
+	resolver.AddCustom("Convolution2DTransposeBias", mediapipe::tflite_operations::RegisterConvolution2DTransposeBias());
+	tflite::InterpreterBuilder builder(*ctx.model, resolver);
+	builder(&ctx.interpreter);
+	TFLITE_MINIMAL_CHECK(ctx.interpreter != nullptr);
```
BenBE, line 131: Possibly something that should be model-specific initialization.

lib/libbackscrub.cc
```diff
@@ -0,0 +1,270 @@
...
+	// clear internal context if present
+	if (info.backscrub_ctx) {
+		backscrub_ctx_t &ctx = *((backscrub_ctx_t *)info.backscrub_ctx);
+		// drop interpreter
+		ctx.interpreter.reset();
+		// drop model
+		ctx.model.reset();
+		// drop context
+		delete &ctx;
+		info.backscrub_ctx = nullptr;
```
BenBE, line 175: Race condition with potential UAF. Always first clear externally visible references, then release associated memory.

BenBE (changes requested):

Makefile
phlash: Good point on parallel builds, adopting your suggestion :smile:

Makefile
phlash: I did this originally to avoid having to pass transitive dependencies back to projects that use `libbackscrub.a`, not to fix anything in _this project_. However, since I think the `make` build herein is never going to be used for other projects, and the `cmake` build can export dependencies then I'll remove this hack as suggested.

lib/libbackscrub.cc
phlash: Another import from the original source, agreed not a helpful comment.

lib/libbackscrub.cc
phlash: Heh, order of implementation bug, I'll go find any other places this may have occurred too.

lib/libbackscrub.h
phlash: I honestly don't like this macro at all - better to return appropriate codes from a library than ditch the whole program, I'll look at a better way of doing this.

lib/libbackscrub.h
phlash: I started on that route, then decided to fix one thing at a time, so yep, C compatible API, which will be much easier when all this implementation detail is hidden away in the library, not shared with callers..

lib/libbackscrub.h
phlash: Yep, as above C-compatible API possible once more state is hidden, although we are presently passing `cv::Mat` objects for raw video frames, and may need to use the plain C alternative from OpenCV...

lib/libbackscrub.h
phlash: Agreed, I think I just missed this :smile:

lib/libbackscrub.cc
phlash: Agreed, and with an API change to hide the implementation detail this should go away too.

lib/libbackscrub.cc
phlash: Possibly, but harmless if the model doesn't use it.

lib/libbackscrub.cc
phlash: Agreed, and another case of removing the implementation detail from callers first, then fixing what remains.

lib/libbackscrub.cc
phlash: Removed entirely, debug filtering is the job of library caller if required.

lib/libbackscrub.h
phlash: Removed entirely.

lib/libbackscrub.h
phlash: All gone, but we may have a problem trying to expose a plain C API:
https://stackoverflow.com/questions/48365991/is-there-a-working-c-interface-for-opencv-3-x
looks like OpenCV no longer has a C API we can use to transport image data...

lib/libbackscrub.h
phlash: As above, no C compatible replacement exists for `cv::Mat` objects:
https://stackoverflow.com/questions/48365991/is-there-a-working-c-interface-for-opencv-3-x

BenBE: okay, will have to refactor this part later on (got some idea on how to get this down to plain old C). We just should keep this in mind for now to reduce the work on this subject later on.

lib/libbackscrub.h
BenBE: I think we can do a somewhat hybrid API: Most functions can be exposed as wrappers with C compatibility and some direct C++ API when C++ is supported. In the worst case this just means, that some C++ API may be hidden for plain C usage, which is not great, but allows for easy fallback in such usecases. But mor on this in a proper issue.

lib/libbackscrub.cc
BenBE: More of an architectural note anyways. ;-)

lib/libbackscrub.cc
```diff
@@ -99,94 +122,125 @@ static normalization_t get_normalization(modeltype_t type) {
...
+		return nullptr;
 	}
 	// Build the interpreter
 	tflite::ops::builtin::BuiltinOpResolver resolver;
 	// custom op for Google Meet network
 	resolver.AddCustom("Convolution2DTransposeBias", mediapipe::tflite_operations::RegisterConvolution2DTransposeBias());
 	tflite::InterpreterBuilder builder(*ctx.model, resolver);
 	builder(&ctx.interpreter);
-	TFLITE_MINIMAL_CHECK(ctx.interpreter != nullptr);
+	if (nullptr == ctx.interpreter) {
```
BenBE, line 180:
```suggestion
	if (!ctx.interpreter) {
```

lib/libbackscrub.cc
```diff
@@ -99,94 +122,125 @@ static normalization_t get_normalization(modeltype_t type) {
...
+	ctx.ondebug = ondebug;
+	ctx.onprep = onprep;
+	ctx.oninfer = oninfer;
+	ctx.onmask = onmask;
+	ctx.caller_ctx = caller_ctx;
 	// Load model
-	ctx.model = tflite::FlatBufferModel::BuildFromFile(info.modelname);
-	TFLITE_MINIMAL_CHECK(ctx.model != nullptr);
+	ctx.model = tflite::FlatBufferModel::BuildFromFile(modelname);
+	if (nullptr == ctx.model) {
```
BenBE, line 161:
```suggestion
	if (!ctx.model) {
```

lib/libbackscrub.cc
```diff
@@ -99,94 +122,125 @@ static normalization_t get_normalization(modeltype_t type) {
...
+	void (*ondebug)(void *ctx, const char *fmt, va_list ap),
+	// ..after preparing video frame
+	void (*onprep)(void *ctx),
+	// ..after running inference
+	void (*oninfer)(void *ctx),
+	// ..after generating mask
+	void (*onmask)(void *ctx),
+	// ..the returned context
+	void *caller_ctx
+	) {
```
BenBE, line 148: Maybe skip the indentation for this line to make the end of the argument list visually more obvious.
```suggestion
) {
```

lib/libbackscrub.cc
```diff
@@ -99,94 +122,125 @@ static normalization_t get_normalization(modeltype_t type) {
...
-		info.backscrub_ctx = nullptr;
-	}
+	delete &ctx;
 }
 
-int calc_mask(calcinfo_t &info) {
-	// Ensure we have a context from init_tensorflow()
-	TFLITE_MINIMAL_CHECK(info.backscrub_ctx!=NULL);
-	backscrub_ctx_t &ctx = *((backscrub_ctx_t *)info.backscrub_ctx);
+int bs_maskgen_process(void *context, cv::Mat &frame, cv::Mat &mask) {
```
BenBE, line 233: Should return `bool` instead if the return value is just used as an success indicator.

lib/libbackscrub.cc
```diff
@@ -99,94 +122,125 @@ static normalization_t get_normalization(modeltype_t type) {
...
+	if (nullptr == context)
```
BenBE, line 234:
```suggestion
	if (!context)
```

lib/libbackscrub.h
```diff
@@ -2,53 +2,38 @@
...
+// Return a new (opaque) mask generation context
+extern void *bs_maskgen_new(
+	// Required parameters
+	const char *modelname,
+	size_t threads,
+	size_t width,
+	size_t height,
+	// Optional (nullable) callbacks with caller-provided context
+	// ..debug output
+	void (*ondebug)(void *ctx, const char *fmt, va_list ap),
```
BenBE, line 22: Variable argument lists make for bad FFI compat. Accepted for now.

BenBE (changes requested):
We're getting there …

lib/libbackscrub.h
phlash: Yep, let's take this to a new issue :)

lib/libbackscrub.cc
phlash: Yep, and possibly something to come back to if model metadata listed operations used (better, provided code to run!)

lib/libbackscrub.cc
phlash: Agreed, I usually rely on truthy-ness of pointers too, this was a result of translating the macro calls, will fix through the code.

lib/libbackscrub.cc
phlash: as above - will fix.

lib/libbackscrub.cc
phlash: Gah, VScode missed one :smile: will fix.

lib/libbackscrub.cc
phlash: Agreed, may need to be revisited as/when a C API appears (which would need to use C99 `stdbool.h` include)

lib/libbackscrub.cc
phlash: ditto earlier - will fix.

lib/libbackscrub.h
phlash: Good point, would ellipsis be any better?

BenBE: Actually worse as not every language supports them. Best is to have a fixed interface with just plain data types, structs or pointers. va_list is somewhat dependent on the ABI of the compiler, as is the ellipsis for varargs. But as said, let's leave this for now and see what features we actually need in the debug calls later on.

phlash: Agreed, formatting within lib.

phlash: Leaving this for reference as/when we want to port off of Linux:
https://stackoverflow.com/questions/40159892/using-asprintf-on-windows

BenBE (approved):
LGTM AFAIC.

phlash:
@peckto Are you happy for me to merge these changes before your threading work? Impacts are likely as I have removed the `calcinfo_t` structure entirely, re-located source files and re-named the functions that handle mask generation... take a look at the re-located [deepseg.cc](https://github.com/phlash/backscrub/blob/libbackscrub/app/deepseg.cc) in the PR. I was also expecting this to take longer :smile: and merge after you :smile:

BenBE:
@phlash Please rebase this PR, seems to have conflicts with `experimental`.

peckto:
@phlash thats fine for me. Adapting threading to libbackscrub is probably better than the other way around.

phlash:
@BenBE - just tried a merge locally, it seems fine to me?
```bash
phlash@bruce:~/code/backscrub$ git branch -avv
* experimental                   81e8b21 [phlash/experimental] Merge branch 'main' into experimental
  libbackscrub                   2a8d351 [phlash/libbackscrub] Remove stdarg from debug callback, cleanup pointer checks
  main                           c15f92e [origin/main] Normalize input correctly for segm_ / selfie (#77)
  master                         841d5cb [origin/master] add a transparent viewer for overlay
  remotes/origin/HEAD            -> origin/main
  remotes/origin/experimental    81e8b21 Merge branch 'main' into experimental
  remotes/origin/main            c15f92e Normalize input correctly for segm_ / selfie (#77)
  remotes/origin/master          841d5cb add a transparent viewer for overlay
  remotes/phlash/experimental    81e8b21 Merge branch 'main' into experimental
  remotes/phlash/libbackscrub    2a8d351 Remove stdarg from debug callback, cleanup pointer checks

phlash@bruce:~/code/backscrub$ git merge --squash libbackscrub 
Updating 81e8b21..2a8d351
Fast-forward
Squash commit -- not updating HEAD
 .gitignore                                           |   3 +-
 CMakeLists.txt                                       |  92 ++++++++++++++++++++++++++++++++++-------
 Makefile                                             |  43 ++++++++++++++++----
 deepseg.cc => app/deepseg.cc                         | 287 +++++++++++++++++--------------------------------------------------------------------------------------------------------------
 lib/libbackscrub.cc                                  | 337 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 lib/libbackscrub.h                                   |  38 +++++++++++++++++
 transpose_conv_bias.cc => lib/transpose_conv_bias.cc |   0
 transpose_conv_bias.h => lib/transpose_conv_bias.h   |   0
 loopback.cc => videoio/loopback.cc                   |   0
 loopback.h => videoio/loopback.h                     |   0
 10 files changed, 526 insertions(+), 274 deletions(-)
 rename deepseg.cc => app/deepseg.cc (61%)
 create mode 100644 lib/libbackscrub.cc
 create mode 100644 lib/libbackscrub.h
 rename transpose_conv_bias.cc => lib/transpose_conv_bias.cc (100%)
 rename transpose_conv_bias.h => lib/transpose_conv_bias.h (100%)
 rename loopback.cc => videoio/loopback.cc (100%)
 rename loopback.h => videoio/loopback.h (100%)

phlash@bruce:~/code/backscrub$ git status
On branch experimental
Your branch is up to date with 'phlash/experimental'.

Changes to be committed:
  (use "git reset HEAD <file>..." to unstage)

	modified:   .gitignore
	modified:   CMakeLists.txt
	modified:   Makefile
	renamed:    deepseg.cc -> app/deepseg.cc
	new file:   lib/libbackscrub.cc
	new file:   lib/libbackscrub.h
	renamed:    transpose_conv_bias.cc -> lib/transpose_conv_bias.cc
	renamed:    transpose_conv_bias.h -> lib/transpose_conv_bias.h
	renamed:    loopback.cc -> videoio/loopback.cc
	renamed:    loopback.h -> videoio/loopback.h

phlash@bruce:~/code/backscrub$ make clean all
rm -rf bin
mkdir -p bin
g++ lib/libbackscrub.cc -fPIC -Ofast -march=native -fno-trapping-math -fassociative-math -funsafe-math-optimizations -Wall -pthread -D DEEPSEG_VERSION=heads/experimental-0-g81e8b21-dirty -I. -I/usr/include/opencv -I tensorflow -I tensorflow/tensorflow/lite/tools/make/downloads/absl -I tensorflow/tensorflow/lite/tools/make/downloads/flatbuffers/include -ggdb -c -o bin/libbackscrub.o
g++ lib/transpose_conv_bias.cc -fPIC -Ofast -march=native -fno-trapping-math -fassociative-math -funsafe-math-optimizations -Wall -pthread -D DEEPSEG_VERSION=heads/experimental-0-g81e8b21-dirty -I. -I/usr/include/opencv -I tensorflow -I tensorflow/tensorflow/lite/tools/make/downloads/absl -I tensorflow/tensorflow/lite/tools/make/downloads/flatbuffers/include -ggdb -c -o bin/transpose_conv_bias.o
ar rv bin/libbackscrub.a bin/libbackscrub.o bin/transpose_conv_bias.o
ar: creating bin/libbackscrub.a
a - bin/libbackscrub.o
a - bin/transpose_conv_bias.o
g++ videoio/loopback.cc -fPIC -Ofast -march=native -fno-trapping-math -fassociative-math -funsafe-math-optimizations -Wall -pthread -D DEEPSEG_VERSION=heads/experimental-0-g81e8b21-dirty -I. -I/usr/include/opencv -I tensorflow -I tensorflow/tensorflow/lite/tools/make/downloads/absl -I tensorflow/tensorflow/lite/tools/make/downloads/flatbuffers/include -ggdb -c -o bin/loopback.o
ar rv bin/libvideoio.a bin/loopback.o
ar: creating bin/libvideoio.a
a - bin/loopback.o
g++ app/deepseg.cc bin/libbackscrub.a bin/libvideoio.a tensorflow/tensorflow/lite/tools/make/gen/linux_x86_64/lib/libtensorflow-lite.a -fPIC -Ofast -march=native -fno-trapping-math -fassociative-math -funsafe-math-optimizations -Wall -pthread -D DEEPSEG_VERSION=heads/experimental-0-g81e8b21-dirty -I. -I/usr/include/opencv -I tensorflow -I tensorflow/tensorflow/lite/tools/make/downloads/absl -I tensorflow/tensorflow/lite/tools/make/downloads/flatbuffers/include -ggdb -lrt -ldl -lopencv_shape -lopencv_stitching -lopencv_superres -lopencv_videostab -lopencv_aruco -lopencv_bgsegm -lopencv_bioinspired -lopencv_ccalib -lopencv_datasets -lopencv_dpm -lopencv_face -lopencv_freetype -lopencv_fuzzy -lopencv_hdf -lopencv_line_descriptor -lopencv_optflow -lopencv_video -lopencv_plot -lopencv_reg -lopencv_saliency -lopencv_stereo -lopencv_structured_light -lopencv_phase_unwrapping -lopencv_rgbd -lopencv_viz -lopencv_surface_matching -lopencv_text -lopencv_ximgproc -lopencv_calib3d -lopencv_features2d -lopencv_flann -lopencv_xobjdetect -lopencv_objdetect -lopencv_ml -lopencv_xphoto -lopencv_highgui -lopencv_videoio -lopencv_imgcodecs -lopencv_photo -lopencv_imgproc -lopencv_core -L tensorflow/tensorflow/lite/tools/make/gen/linux_x86_64/lib -ltensorflow-lite -ldl -o bin/deepseg

phlash@bruce:~/code/backscrub$
```

Issue #92, opened 2021-06-17
Does this work in Windows?


tdbe:
Never heard of v4l2loopback . Is ithard to make this thing work in windows? Because I want to be able to just go to a computer, any computer, which 99% of the time is Windows, and run this :)

BenBE:
`v4l2loopback` is a kernel mode driver for Linux. No chance porting that driver to Windows.
The part that may run on Windows is the actual image processing part inside `libbackscrub`, because that is independent of the driver interface. But the library is only the processing, the virtual cam part will need a completely different approach on Windows.

phlash:
Hopefully someone will get time to port the core library to Windows, along with my OBS plugin that uses it, that is the reason we have split it out :smile:

BenBE:
@phlash The core library part should mostly run on Windows as is (with maybe minor fixes).

Where I currently see the biggest open questions is with the Camera Interface. Once we have image data for processing, we should be fine even on Windows, but pushing the processed image into some camera device may be tricky.

OmarJay1:
I've done single image replacement in Windows, although there seems to be issues with the models.  Probably my mucking with the code to make it build in Windows.

One issue with Windows is pthreads, which Windows doesn't have natively.  There are 2 packages that seem like they'll work:

http://www.sourceware.org/pthreads-win32/
https://github.com/BrianGladman/pthreads

I don't think video processing would be an issue if it's handled inside an app for the Windows build instead of the loopback driver.  

I can make a simple mockup and post as a PR if there are no objections to this approach.  

Thanks.

phlash:
Thanks for digging in @OmarJay1! I presume you are attempting to build the `experimental` branch, including the underlying TFLite with CMake and MSVC (from Visual Studio)?

This ticket (https://github.com/tensorflow/tensorflow/issues/47166) seems to indicate TFLite builds work for x64 binaries from v2.4.1 tagged Tensorflow source..

I might have a go with a cross-compiler (mingw64) locally myself..

OmarJay1:
I got something to build and with the default settings it seems to work.  It's very messy at this point.  I'll try to clean it up a bit.

Build-wise the main issues were pthreadpool, which TFLite needs, and pthreads which deepseg.cc needs.  I had to get the latest version of pthreadpool and build against that.  

For pthreads, I hacked in overrides to the functions using std::thread and std::mutex.  

[...]

PR #94, opened 2021-07-02
Initialise Tensorflow submodule to correct version in CMake builds

phlash:
As noted in #92, our CMake builds don't pull Tensorflow source like the Makefile does, this fixes it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,10 +1,13 @@
 cmake_minimum_required(VERSION 3.16)
 project(backscrub CXX)
 
+# allow override of Tensorflow location
 if(NOT DEFINED TENSORFLOW)
   set(TENSORFLOW tensorflow)
+  set(TF_CHECKOUT true)
 endif()
 
+# find Git and use it to get backscrub version
 find_package(Git)
 if(GIT_FOUND)
   execute_process(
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -1,10 +1,13 @@
...
+# allow override of Tensorflow location
 if(NOT DEFINED TENSORFLOW)
   set(TENSORFLOW tensorflow)
+  set(TF_CHECKOUT true)
 endif()
 
+# find Git and use it to get backscrub version
 find_package(Git)
 if(GIT_FOUND)
   execute_process(
```
BenBE, line 13: IMHO `git describe --long --tags --always --dirty` gives more descriptive results …

BenBE (approved):
LGTM.

CMakeLists.txt
phlash: You may be right, I'm not really fussed.. I tried both on current main and experimental branches:
```bash
% git checkout experimental
...
% git describe --long --always --dirty --all
heads/experimental-0-gaa243fc
% git describe --long --always --dirty --tags
v0.3.0-7-gaa243fc
% git checkout main
...
% git describe --long --always --dirty --all
heads/main-0-g7ec492c
% git describe --long --always --dirty --tags
7ec492c
```
I would note that this is not changed by this PR :smile:

PR #95, opened 2021-08-05
Correct timing details, simplify thread synchronisation

phlash:
Minor fixes to correct the mask processing loop timing & FPS value, and to avoid order-dependency of class variables when starting processing thread.

### DIFF
--- a/app/deepseg.cc
+++ b/app/deepseg.cc
@@ -138,9 +138,10 @@ static bool is_number(const std::string &s) {
 
 class CalcMask final {
 protected:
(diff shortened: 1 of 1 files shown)


app/deepseg.cc
```diff
@@ -266,6 +246,10 @@ class CalcMask final {
 	}
 };
 
+static bool is_number(const std::string &s) {
+  return !s.empty() && std::all_of(s.begin(), s.end(), ::isdigit);
```
BenBE, line 250: Indentation

BenBE (approved):
Minor issue, otherwise LGTM.

PR #96, opened 2021-08-07
Experimental merge to main

phlash:
As discussed, we now have no PRs against `experimental` which has a number of features that improve on `main`, such that it would benefit from being merged...

User-visible features:
 * background and foreground are alpha blended to smooth mask edges (partial 'light wrapping')
 * background can be a blurred form or the original rather than a separate image
 * image segmentation runs on a separate thread to maintain video rate throughput

Developer-visible features:
 * all threading uses C++11 `std::thread` package
 * builds are labelled from git version
 * codebase is now split into a library that performs image segmentation/mask generation and a wrapper app that collects inputs, applies the mask and blends outputs. This permits the library to be re-used in other context (such as an OBS plugin)
 * CMake builds now support installation correctly, and checkout correct version of tensorflow
 * OpenCV API usage updated to avoid older C constants in favour of current namespaced C++ values


AUTHORS
```diff
@@ -0,0 +1,13 @@
...
+In surname alphabetical order, the following people have also contributed to
+deepbacksub:
+
+  Phil Ashby <phil.github@ashbysoft.com>
+  Ismael Barros² <razielmine@gmail.com>
+  Benny Baumann <BenBE@geshi.org>
+  Cristian Ciocaltea <cristian.ciocaltea@gmail.com>
+  Mark Kogan <mkogan@redhat.com>
+  oleid <oleid@users.noreply.github.com>
+  Tobias Specht <specht.tobias@gmx.de>
```
BenBE, line 13: Please add LF @ EOF

AUTHORS
```diff
@@ -0,0 +1,13 @@
+Backscrub (previously Deepbacksub) was conceived of and implemented by
+Florian Echtler <floe@butterbrot.org> in early 2020.
+
+In surname alphabetical order, the following people have also contributed to
+deepbacksub:
+
+  Phil Ashby <phil.github@ashbysoft.com>
+  Ismael Barros² <razielmine@gmail.com>
```
BenBE, line 8: Is the `²` a footnote? Where's it specified?

CMakeLists.txt
```diff
@@ -1,8 +1,42 @@
...
+  set(TF_CHECKOUT true)
+endif()
+
+# find Git and use it to get backscrub version
+find_package(Git)
+if(GIT_FOUND)
+  execute_process(
+    COMMAND ${GIT_EXECUTABLE} describe --all --long --always --dirty
+       OUTPUT_STRIP_TRAILING_WHITESPACE
+       OUTPUT_VARIABLE DEEPSEG_VERSION)
```
BenBE, line 16: Consistent indentation depth of 2 per level?

AUTHORS
phlash: Not a footnote, it's their current Github name.

AUTHORS
phlash: Doh.. yep :smile:

CMakeLists.txt
phlash: OK, have also fixed indentation on install target.

phlash:
Thinking we should probably lose `EXPERIMENTAL.md` as these enhancements better managed through issues?

BenBE (approved):

phlash:
Cool, thanks @BenBE, I'll give @floe a day or so to chip in with any possible blocker before merging this.

floe (approved):
No further suggestions from my side, merge away :-) Thanks for all your work!

PR #97, opened 2021-08-08
Windows build of libbackscrub

phlash:
Omit wrapper application from CMake build (it depends on video4linux)
Revert to older structure initialisation (pre c++17) to support MSVC compiler


phlash:
As per #92, this 'works for me' on a fresh Windows Server 2016/9 VM with Visual Studio 2019 Community Edition in Azure. I would appreciate more testing on other development machines :grin:

BenBE (approved):

Issue #70, opened 2021-04-15
Recommendation: ability to change background while running.


MartinKlevs:
Restarting the process could be inconvenient in some scenarios. The simplest way to do this would be to regularly check wether the background file has been replaced.

BenBE:
If you need the ability to change the background at any time it's best to use a tool like OBS Studio and use deepseg only to perform the background segmentation. That's also what the <kbd>s</kbd> shortcut on the debug window is for, so you can disable the background replacement at any time when another tool is providing the background for you instead.

MartinKlevs:
This would work fine for a pixel-sharp segmentation, however I am experimenting with smooth mask borders and a smooth transition is hard to fix properly in OBS. One could also output just the mask and mix it with the webcam in OBS, however that can lead to problems with multiple programs accesing the webcam.

Besides that, it would also be a good convenience feature, that uses less resources, than mixing with OBS.

phlash:
Possibly related to #43 where local file input / video stream backgrounds are discussed, tagging as an experiment to try..

Issue #43, opened 2021-02-11
Recommendation: Release this to work with video file input and output


richardpickett:
It'd be nice to feed an already-recorded video through this to remove the background.

BenBE:
While the current code structure is quite fixed on reading the video stream from the V4L video device it shouldn't be too hard, implementing a different processing mode for the input/output, once Something like #38 and #41 are implemented (amongst some other code structure changes).

phlash:
This is a similar need to the one I had for animated backgrounds, the separated input module I have (https://github.com/phlash/deepbacksub/blob/65c7f4a8f0f5ddd1cfc6305fef825690fc1a1931/capture.cc#L59) allows local files to be used as source material for either input stream. This supports @BenBE's statement that it's not too hard, and for me points towards doing one thing well (segmentation and mask generation), with other features (background substitution, offline processing, etc.) perhaps delegated to existing machinery (eg: ffmpeg, obs)?

PR #100, opened 2021-08-18
New module for live backgrounds

phlash:
New code module to handle backgrounds, and in particular live backgrounds from local files or network streams.

Partly addresses #70, in that backgrounds are now a more flexible item, although no UI is present to select one while running.
Also partly addresses #43, in that video backgrounds are now possible.

### DIFF
--- a/app/background.cc
+++ b/app/background.cc
@@ -1,0 +1,161 @@
+/* This is licensed software, @see LICENSE file.
+ * Authors - @see AUTHORS file. */
+
+#include <stdio.h>
+#include <time.h>
+#include <unistd.h>
+#include <thread>
+#include <mutex>
+#include <opencv2/videoio.hpp>
+#include <opencv2/imgproc.hpp>
+#include <opencv2/highgui.hpp>
+
+// Internal state of background processing
+struct background_t {
+    int dbg;
+    bool vid;
+    volatile bool run;
+    cv::VideoCapture cap;
+    int frm;
+    double fps;
+    cv::Mat raw;
+    std::thread thr;
+    std::mutex mux;
+};
+
+// Internal video reader thread
+static void read_thread(background_t *pbkd) {
+    if (pbkd->dbg) fprintf(stderr, "background: thread start\n");
+    struct timespec last;
+    struct timespec proc;
+    clock_gettime(CLOCK_REALTIME, &last);
+    proc = last;
+    while (pbkd->run) {
+        // read frame under lock
+        pbkd->mux.lock();
+        if (pbkd->cap.read(pbkd->raw)) {
+            pbkd->frm += 1;
+            // grab timing point and calculate frame period
+            struct timespec now;
+            clock_gettime(CLOCK_REALTIME, &now);
+            double nsec = (double)(now.tv_sec-last.tv_sec)*1e9 + (double)(now.tv_nsec-last.tv_nsec);
+            last = now;
+            // display thumbnail frame with overlay info if debug enabled
+            if (pbkd->dbg) {
+                char msg[40];
+                snprintf(msg, sizeof(msg), "FPS:%0.1f FRM:%d", 1e9/nsec, pbkd->frm);
+                cv::Mat frame;
+                cv::resize(pbkd->raw, frame, cv::Size(240,160));
+                cv::putText(frame, msg, cv::Point(5,frame.rows-5), cv::FONT_HERSHEY_PLAIN, 1.0, cv::Scalar(255,255,255));
+                cv::imshow("Background", frame);
+            }
+            pbkd->mux.unlock();
+            // wait for next frame, some sources are real-time, others are not, this ensures all play in real-time.
+            if (pbkd->frm>1) {
+                long adj = (now.tv_sec-proc.tv_sec)*1000000000l + (now.tv_nsec-proc.tv_nsec);
+                now.tv_sec = 0;
+                now.tv_nsec= (long)(1.0/pbkd->fps*1e9);
+                // adjust for processing time or skip if less than that
+                if (now.tv_nsec > adj) {
+                    now.tv_nsec -= adj;
+                    nanosleep(&now, nullptr);
+                } else {
+                    if (pbkd->dbg) fprintf(stderr, "background: wait=%ld adj=%ld\n", now.tv_nsec, adj);
+                }
+                clock_gettime(CLOCK_REALTIME, &proc);
+            }
+        } else {
+            // no more frames, but if we processed some, try to reset position and go again
+            if (pbkd->frm>0 && pbkd->cap.set(cv::CAP_PROP_POS_FRAMES, 0)) {
+                std::unique_lock<std::mutex> hold(pbkd->mux);
+                pbkd->frm = 0;
+            } else {
+                // unable to reset or previous attempt produced no more frames - stop
+                if (pbkd->dbg) fprintf(stderr, "background: thread stopping at end of stream and not resettable\n");
+                break;
+            }
+        }
+    }
+    if (pbkd->dbg) fprintf(stderr, "background: thread stop\n");
+}
+
+void *load_background(const char *path, int debug) {
+    background_t *pbkd = new background_t;
+    background_t &bkd = *pbkd;
+
+    bkd.dbg = debug;
+    bkd.cap.open(path, cv::CAP_ANY);    // explicitly ask for auto-detection of backend
+    if (!bkd.cap.isOpened()) {
+        if (bkd.dbg) fprintf(stderr, "cv::VideoCapture cannot open: %s\n", path);
+        delete pbkd;
+        return nullptr;
+    }
+    bkd.cap.set(cv::CAP_PROP_CONVERT_RGB, true);
+    int fcc = (int)bkd.cap.get(cv::CAP_PROP_FOURCC);
+    bkd.fps = bkd.cap.get(cv::CAP_PROP_FPS);
+    int cnt = (int)bkd.cap.get(cv::CAP_PROP_FRAME_COUNT);
+    // Here be the logic...
+    //  if: can read 2 video frames => it's a video
+    //  else: is loaded as an image => it's an image
+    //  else: it's not usable.
+    if (bkd.cap.read(bkd.raw) && bkd.cap.read(bkd.raw)) {
+        // it's a video, try a reset and start reader thread..
+        if (bkd.cap.set(cv::CAP_PROP_POS_FRAMES, 0))
+            bkd.frm = 0;
+        else
+            bkd.frm = 2;
+        bkd.vid = true;
+        bkd.run = true;
+        bkd.thr = std::thread(read_thread, pbkd);
+    } else {
+        // static image file, try loading..
+        bkd.cap.release();
+        bkd.raw = cv::imread(path);
+        bkd.vid = false;
+        if (bkd.raw.empty()) {
+            if (bkd.dbg) fprintf(stderr, "cv::imread cannot read: %s\n", path);
+            delete pbkd;
+            return nullptr;
+        }
+    }
+    if (bkd.dbg) fprintf(stderr, "background properties:\n\tvid: %s\n\tfcc: %08x (%.4s)\n\tfps: %f\n\tcnt: %d\n",
+        bkd.vid ? "yes":"no", fcc, (char *)&fcc, bkd.fps, cnt);
+    return pbkd;
+}
+
+int grab_background(void *handle, int width, int height, cv::Mat& out) {
+    background_t *pbkd = (background_t *)handle;
+    if (!pbkd)
+        return -1;
+    // static image or video?
+    int frm ;
+    if (pbkd->vid) {
+        // grab frame & frame no. under mutex
+        std::unique_lock<std::mutex> hold(pbkd->mux);
+        cv::resize(pbkd->raw, out, cv::Size(width, height));
+        frm = pbkd->frm;
+    } else {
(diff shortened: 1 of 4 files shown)


app/background.cc
```diff
@@ -0,0 +1,161 @@
...
+    std::thread thr;
+    std::mutex mux;
+};
+
+// Internal video reader thread
+static void read_thread(background_t *pbkd) {
+    if (pbkd->dbg) fprintf(stderr, "background: thread start\n");
+    struct timespec last;
+    struct timespec proc;
+    clock_gettime(CLOCK_REALTIME, &last);
```
BenBE, line 31: std::chrono -> monotonic_clock

app/background.cc
```diff
@@ -0,0 +1,161 @@
...
+    proc = last;
+    while (pbkd->run) {
+        // read frame under lock
+        pbkd->mux.lock();
```
BenBE, line 35: Never unlocked if read feils

BenBE: Also use RAII locks via std::lock_guard here if possible.

app/background.cc
```diff
@@ -0,0 +1,161 @@
...
+                cv::resize(pbkd->raw, frame, cv::Size(240,160));
+                cv::putText(frame, msg, cv::Point(5,frame.rows-5), cv::FONT_HERSHEY_PLAIN, 1.0, cv::Scalar(255,255,255));
+                cv::imshow("Background", frame);
+            }
+            pbkd->mux.unlock();
+            // wait for next frame, some sources are real-time, others are not, this ensures all play in real-time.
+            if (pbkd->frm>1) {
+                long adj = (now.tv_sec-proc.tv_sec)*1000000000l + (now.tv_nsec-proc.tv_nsec);
+                now.tv_sec = 0;
+                now.tv_nsec= (long)(1.0/pbkd->fps*1e9);
```
BenBE, line 57:
```suggestion
                now.tv_nsec= (long)(1.0e9/pbkd->fps);
```

app/background.cc
```diff
@@ -0,0 +1,161 @@
...
+
+void *load_background(const char *path, int debug) {
+    background_t *pbkd = new background_t;
+    background_t &bkd = *pbkd;
+
+    bkd.dbg = debug;
+    bkd.cap.open(path, cv::CAP_ANY);    // explicitly ask for auto-detection of backend
+    if (!bkd.cap.isOpened()) {
+        if (bkd.dbg) fprintf(stderr, "cv::VideoCapture cannot open: %s\n", path);
+        delete pbkd;
```
BenBE, line 90: Use try/catch to ensure memory is freed on error

app/background.cc
```diff
@@ -0,0 +1,161 @@
...
+    // Here be the logic...
+    //  if: can read 2 video frames => it's a video
+    //  else: is loaded as an image => it's an image
+    //  else: it's not usable.
+    if (bkd.cap.read(bkd.raw) && bkd.cap.read(bkd.raw)) {
+        // it's a video, try a reset and start reader thread..
+        if (bkd.cap.set(cv::CAP_PROP_POS_FRAMES, 0))
+            bkd.frm = 0;
+        else
+            bkd.frm = 2;
```
BenBE, line 106:
```suggestion
        bkd.frm = bkd.cap.set(cv::CAP_PROP_POS_FRAMES, 0) ? 0 : 2;
```

Apart from this explain those magic numbers …

app/background.cc
```diff
@@ -0,0 +1,161 @@
...
+    } else {
+        // resize still image as requested into out
+        cv::resize(pbkd->raw, out, cv::Size(width, height));
+        frm = 1;
+    }
+    return frm;
+}
+
+void drop_background(void *handle) {
+    if (handle) {
```
BenBE, line 146: Negate condition to exit early. Reduces indentation for the proc overall.

app/background.h
```diff
@@ -0,0 +1,21 @@
+/* This is licensed software, @see LICENSE file.
+ * Authors - @see AUTHORS file. */
+
+#ifndef _BACKGROUND_H_
+#define _BACKGROUND_H_
+
+#include <opencv2/core/mat.hpp>
+
```
BenBE, line 8: Forward-declare `struct background_t;` to pass typed pointers to all those functions. Doesn't affect actual API/structure for this structure on actual definition.

app/deepseg.cc
```diff
@@ -454,18 +458,24 @@ int main(int argc, char* argv[]) try {
...
 			ai.get_output_mask(mask);
 
+			// get background frame
+			if (pbk) {
+				grab_background(pbk, width, height, bg);
+			} else {
+				if (blur_strength) {
+					raw.copyTo(bg);
+					cv::GaussianBlur(bg,bg,cv::Size(blur_strength,blur_strength),0);
+				}
```
BenBE, line 472: Any reason blurring the background is skipped for animated backgrounds? IMHO there's no real reason why this shouldn't be possible.

app/deepseg.cc
```diff
@@ -403,17 +409,15 @@ int main(int argc, char* argv[]) try {
 	printf("back:   %s\n", back ? back : "(none)");
 	printf("model:  %s\n\n", modelname);
 
-	cv::Mat bg;
-	if (back) {
-		bg = cv::imread(back);
-	}
-	if (bg.empty()) {
+	void *pbk = (back) ? load_background(back, debug) : nullptr;
```
BenBE, line 412:
```suggestion
	std::shared_ptr<background_t> pbk(back ? load_background(back, debug) : nullptr, [](background_t* p) {
		if(p)
			drop_background(p);
	});
```

BenBE (changes requested):

app/background.cc
phlash: Yeah - I cut/past that from test code :smile: will fix.

app/background.cc
phlash: Heh, yep, will fix

app/background.cc
phlash: Good point - yep, will fix.

app/background.cc
phlash: I will add a suitable comment!

app/background.cc
phlash: Thanks - yep, will fix.

app/background.h
phlash: Ah yep, will fix.

app/deepseg.cc
phlash: Yep, along with typing as earlier.

app/deepseg.cc
phlash: Good point, I think the original logic was that you have a background exactly as you would like it (still or animated), or you wish to blur the original video background and use that. But possibly a poor assumption, we can blur anything :smile:

app/background.cc
phlash: Yep, spotted that - fell in my own trap! Now using RAII locks.

app/background.cc
BenBE: Especially note the difference between wall clock time and monotonic clock time …

phlash: decided to use `steady_clock` (monotonic)

app/background.cc
```diff
@@ -24,45 +23,45 @@ struct background_t {
...
-                    nanosleep(&now, nullptr);
-                } else {
-                    if (pbkd->dbg) fprintf(stderr, "background: wait=%ld adj=%ld\n", now.tv_nsec, adj);
+                auto adj = std::chrono::duration_cast<std::chrono::nanoseconds>(now-proc);
+                auto del = std::chrono::nanoseconds((long)(1e9/pbkd->fps));
+                // adjust for processing time and sleep or skip if no time left
+                if (del > adj) {
+                    del -= adj;
+                    std::this_thread::sleep_for(del);
                 }
```
BenBE: Prone to overall jitter AFAIC. Especially as the `std::this_thread::sleep_for` does not guarantee exact timings.

app/background.cc
```diff
@@ -79,52 +78,56 @@ static void read_thread(background_t *pbkd) {
...
-    } else {
-        // static image file, try loading..
-        bkd.cap.release();
-        bkd.raw = cv::imread(path);
-        bkd.vid = false;
-        if (bkd.raw.empty()) {
-            if (bkd.dbg) fprintf(stderr, "cv::imread cannot read: %s\n", path);
-            delete pbkd;
+std::shared_ptr<background_t> load_background(const char *path, int debug) {
+    std::shared_ptr<background_t> pbkd(new background_t);
```
BenBE, line 82:
```suggestion
    std::shared_ptr<background_t> pbkd = std::make_shared<background_t>();
```

Even might:
```suggestion
    auto pbkd = std::make_shared<background_t>();
```

As the return type of `std::make_shared<T>` is `std::shared_ptr<T>` …

Also note the proposed dtor trick I commented yesterday.

app/background.cc
```diff
@@ -79,52 +78,56 @@ static void read_thread(background_t *pbkd) {
...
+            pbkd->cap.release();
+            pbkd->raw = cv::imread(path);
+            pbkd->vid = false;
+            if (pbkd->raw.empty()) {
+                if (pbkd->dbg) fprintf(stderr, "background: imread cannot open: %s\n", path);
+                return nullptr;
+            }
+        }
+        if (pbkd->dbg) fprintf(stderr, "background properties:\n\tvid: %s\n\tfcc: %08x (%.4s)\n\tfps: %f\n\tcnt: %d\n",
+            pbkd->vid ? "yes":"no", fcc, (char *)&fcc, pbkd->fps, cnt);
```
BenBE, line 118:
```suggestion
        if (pbkd->dbg)
                fprintf(stderr, "background properties:\n\tvid: %s\n\tfcc: %08x (%.4s)\n\tfps: %f\n\tcnt: %d\n",
                        pbkd->vid ? "yes" : "no", fcc, (char *)&fcc, pbkd->fps, cnt);
```

floe:
General question: would it maybe make sense to use GStreamer instead for the whole video streaming thing? (IIRC you can use GStreamer pipelines as a video capture device in OpenCV, although I'm not 100% sure anymore about how.)

BenBE:
Would be interesting to take the contents of another window as background … Not sure if GStreamer allows for that (but IIRC it somehow did) …

app/background.cc
phlash: Any thoughts on what to use from `std::` to improve the timing? I don't think there are any underlying implementations that provide guarantees on resolution, hence nothing in the std library can either?

app/background.cc
phlash: Ah - I missed the dtor trick, that makes sense now I've re-read it!

phlash:
> General question: would it maybe make sense to use GStreamer instead for the whole video streaming thing? (IIRC you can use GStreamer pipelines as a video capture device in OpenCV, although I'm not 100% sure anymore about how.)

I've looked at GStreamer a little, while considering a module to connect libbackscrub into the ecosystem, it's quite invasive to the programming style (ie: you end up writing a GApplication and supporting paraphernalia), so I wouldn't be keen to adapt this codebase (it would be a re-write), instead continue with creating a module that other GStreamer apps can use.

app/background.cc
phlash: I've introduced some min/max measures, and switched to absolute target time via `std::this_thread::sleep_util()` calculated from frame rate. This doesn't show visible jitter (to me anyway!) with a smoothly moving video (eg: https://file-examples-com.github.io/uploads/2020/03/file_example_WEBM_1920_3_7MB.webm). would still welcome thoughts :smile:

app/background.cc
```diff
@@ -45,23 +46,24 @@ static void read_thread(std::shared_ptr<background_t> pbkd) {
...
-            if (pbkd->frm>1) {
-                auto adj = std::chrono::duration_cast<std::chrono::nanoseconds>(now-proc);
-                auto del = std::chrono::nanoseconds((long)(1e9/pbkd->fps));
-                // adjust for processing time and sleep or skip if no time left
-                if (del > adj) {
-                    del -= adj;
-                    std::this_thread::sleep_for(del);
-                }
-                proc = std::chrono::steady_clock::now();
+            proc += std::chrono::nanoseconds((long)(1e9/pbkd->fps));
```
BenBE, line 63: What about dropping frames when too slow?

app/background.cc
```diff
@@ -78,10 +80,32 @@ static void read_thread(std::shared_ptr<background_t> pbkd) {
...
+        // clean up
+        pbkd->raw.release();
+    }
+    delete pbkd;
+}
+
 std::shared_ptr<background_t> load_background(const char *path, int debug) {
-    std::shared_ptr<background_t> pbkd(new background_t);
+    // allocate a shared pointer around storage for the handle, associate custom deleter to clean up when eventually released
+    auto pbkd = std::shared_ptr<background_t>(new background_t, [](background_t *p) { drop_background(p); });
```
BenBE, line 104:
```suggestion
    auto pbkd = std::shared_ptr<background_t>(new background_t, drop_background });
```

app/background.cc
phlash: Good question - if we run out of cycles simply reading frames.. we have bigger problems? Difficult to drop frames on network streams (especially if they are http-like streams), so non-trivial to do this I think?

app/background.cc
phlash: heh, of course :smile:

BenBE (approved):

phlash:
..and now I remember that the usage text should probably change :smile: bug fix incoming!

Issue #101, opened 2021-08-20
model could not be loaded


DarkVik:
I got everything build and installed , but get an error message that the models/segm_full_v679.tflite could not be opened.

Any idea?

**me@Darkvik:/usr/local/bin$ v4l2-ctl --list-devices**
v4l2loopback (platform:v4l2loopback-002):
        /dev/video2

Integrated Camera: Integrated C (usb-0000:00:14.0-8):
        /dev/video0
        /dev/video1

**me@Darkvik:/usr/local/bin$ backscrub -d -d -c /dev/video0 -v /dev/video2**
backscrub version heads/main-0-g9b22962
(c) 2021 by floe@butterbrot.org & contributors
https://github.com/floe/backscrub
debug:  2
ccam:   /dev/video0
vcam:   /dev/video2
width:  640
height: 480
flip_h: no
flip_v: no
threads:2
back:   (none)
model:  models/segm_full_v679.tflite

vid_format->type                = 2
vid_format->fmt.pix.width       = 640
vid_format->fmt.pix.height      = 480
vid_format->fmt.pix.pixelformat = 1448695129
vid_format->fmt.pix.sizeimage   = 614400
vid_format->fmt.pix.field       = 1
vid_format->fmt.pix.bytesperline= 1280
vid_format->fmt.pix.colorspace  = 8

**ERROR: Could not open 'models/segm_full_v679.tflite'.
error: unable to load model from file: 'models/segm_full_v679.tflite'.**
Error: Could not create mask context

**me@Darkvik:/usr/local/bin$ ls -l /usr/local/share/backscrub/models/segm_full_v679.tflite** 
-rw-r--r-- 1 root root 407248 aug 20 14:34 /usr/local/share/backscrub/models/segm_full_v679.tflite

floe:
Mmh, try giving the full path to backscrub (`... -m /usr/local/share/backscrub/models/segm_full_v679.tflite`).

BenBE:
Maybe we should try to put our models as resources to `/usr/share/backscrub/models` following the FHS … People could still override that path by giving a full path name, but paths without a `/` could be taken to reference that resource directory.

phlash:
Given that a cmake `make install` will put everything in `CMAKE_INSTALL_PREFIX` (typically `/usr/local`), and this should 'just work' as the OP expected (rule of least surprise), I think this comes down to a search order for models (or backgrounds?):

 * Check for a local file with the exact provided path (current behaviour), if not found:
[...]

PR #111, opened 2021-09-10
path resolver to aid in finding image and models

phlash:
As mentioned in #101, this adds a path resolver mechanism to help find/load backgrounds and models.


phlash:
Please note that there are some enhancements from my earlier implementation:
 * ignores anything that looks like a URI schema (using RFC3986 definition)
 * emits both supplied and resolved paths to help with debugging command line usage

CMakeLists.txt
```diff
@@ -131,12 +131,15 @@ export(TARGETS
...
   FILE BackscrubTargets.cmake)
 
-# installation names for library, header and models
+# installation names for library, header, backgrounds and models
 if(NOT WIN32)
 install(TARGETS deepseg)
 endif()
 install(TARGETS backscrub)
 install(FILES lib/libbackscrub.h DESTINATION include/backscrub)
+install(DIRECTORY images
```
phlash, line 140: Perhaps rename to `backgrounds` now?

app/deepseg.cc
```diff
@@ -256,6 +263,57 @@ static bool is_number(const std::string &s) {
...
+	std::string result;
+	// Check for network (URI) schema and return as-is
+	// https://www.rfc-editor.org/rfc/rfc3986#section-3.1
+	// however we require at least two chars in the scheme to allow driver letters to work on Windows..
+	if (std::regex_match(provided, std::regex("^[[:alpha:]][[:alnum:]+-.]{1,}:.*$")))
+		return provided;
+	// We use std::ifstream to check we can open each test path read-only, in order:
+	// 1. exactly what was provided
+	if (std::ifstream(provided).good())
+		return provided;
```
phlash, line 276: As yet - this doesn't stop if the `provided` path contains a '/' (as suggested by @BenBE in #101), so it will look for sub-trees in other locations.. useful or confusing?

app/deepseg.cc
```diff
@@ -268,15 +326,15 @@ int main(int argc, char* argv[]) try {
 	size_t threads= 2;
 	size_t width  = 640;
 	size_t height = 480;
-	const char *back = nullptr; // "images/background.png";
-	const char *vcam = "/dev/video0";
-	const char *ccam = "/dev/video1";
+	const char *back = nullptr;
+	const char *vcam = "/dev/video1";
+	const char *ccam = "/dev/video0";
 	bool flipHorizontal = false;
```
phlash, line 332: We might want to be more intelligent in defaulting camera paths too (ie: search for them)?

app/deepseg.cc
```diff
@@ -268,15 +326,15 @@ int main(int argc, char* argv[]) try {
 	size_t threads= 2;
 	size_t width  = 640;
 	size_t height = 480;
-	const char *back = nullptr; // "images/background.png";
-	const char *vcam = "/dev/video0";
-	const char *ccam = "/dev/video1";
+	const char *back = nullptr;
+	const char *vcam = "/dev/video1";
+	const char *ccam = "/dev/video0";
```
phlash, line 331: We might want to be more intelligent in defaulting camera paths too (ie: search for them)?

app/deepseg.cc
```diff
@@ -268,15 +326,15 @@ int main(int argc, char* argv[]) try {
...
 	bool flipHorizontal = false;
 	bool flipVertical   = false;
 	int fourcc = 0;
 	size_t blur_strength = 0;
 
-	const char* modelname = "models/selfiesegmentation_mlkit-256x256-2021_01_19-v1215.f16.tflite";
+	const char* modelname = "selfiesegmentation_mlkit-256x256-2021_01_19-v1215.f16.tflite";
```
phlash, line 337: Default is now the installed model - developers need to supply a path or install a build... good or bad?

app/deepseg.cc
BenBE: The idea behind skipping the search for sub-trees was to a) mimic the behavior when looking for programs in the PATH and b) to allow for relative paths to always work as expected.

app/deepseg.cc
BenBE: You mean like a path resolution that accepts `video1` in place of `/dev/video1`? I.e. transparently do path lookup using `/dev` for filename-only names?

app/deepseg.cc
```diff
@@ -410,18 +470,24 @@ int main(int argc, char* argv[]) try {
...
+	}
 
 	// Create debug window early (ensures highgui is correctly initialised on this thread)
 	if (debug > 1) {
 		cv::namedWindow(DEBUG_WIN_NAME, cv::WINDOW_AUTOSIZE | cv::WINDOW_GUI_EXPANDED);
 	}
 
 	// Load background if specified
-	auto pbk((back) ? load_background(back, debug) : nullptr);
+	auto pbk((s_backg.empty()) ? nullptr : load_background(s_backg.c_str(), debug));
```
BenBE, line 488:
```suggestion
	auto pbk(s_backg.empty() ? nullptr : load_background(s_backg.c_str(), debug));
```

app/deepseg.cc
```diff
@@ -448,7 +514,7 @@ int main(int argc, char* argv[]) try {
 
 	cv::Mat mask(height, width, CV_8U);
 	cv::Mat raw;
-	CalcMask ai(modelname, threads, width, height);
+	CalcMask ai(s_model.c_str(), threads, width, height);
```
BenBE, line 517: Sure that we can't avoid passing raw pointers around?

BenBE (changes requested):

CMakeLists.txt
phlash: OK, will do.

app/deepseg.cc
phlash: OK, emulating PATH search makes intuitive sense, I'm now wondering if we should support a `BACKSCRUB_PATH` ?

app/deepseg.cc
phlash: I was thinking of assuming the first video device is the physical camera (typically `/dev/video0`), and the last is the virtual camera, using a similar scanning technique to OBS (which enumerates `/sys/class/video4linux`)

app/deepseg.cc
```diff
@@ -256,6 +263,57 @@ static bool is_number(const std::string &s) {
 	return !s.empty() && std::all_of(s.begin(), s.end(), ::isdigit);
 }
 
+std::string resolve_path(const std::string& provided, const std::string& type) {
+	std::string result;
+	// Check for network (URI) schema and return as-is
+	// https://www.rfc-editor.org/rfc/rfc3986#section-3.1
+	// however we require at least two chars in the scheme to allow driver letters to work on Windows..
+	if (std::regex_match(provided, std::regex("^[[:alpha:]][[:alnum:]+-.]{1,}:.*$")))
```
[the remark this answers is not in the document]
phlash: What would we gain from CTRE (apart from a tiny speed up during startup)? Is it worth the pain of another dependency? Or were you thinking about target-architecture specific RE's (given my comment about drive letter!)?

app/deepseg.cc
```diff
@@ -256,6 +263,57 @@ static bool is_number(const std::string &s) {
...
+	// We use std::ifstream to check we can open each test path read-only, in order:
+	// 1. exactly what was provided
+	if (std::ifstream(provided).good())
+		return provided;
+	// 2. prefixed with compile-time install path
+	result = _STR(INSTALL_PREFIX);
+	result.append("/share/backscrub/");
+	result.append(type);
+	result.append("/");
+	result.append(provided);
```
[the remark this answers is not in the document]
phlash: Doh - yep, I was obviously working through piece-by-piece when developing :smile:

app/deepseg.cc
```diff
@@ -256,6 +263,57 @@ static bool is_number(const std::string &s) {
...
+		binloc[n] = 0;
+		result = binloc;
+		size_t pos = result.rfind('/');
+		pos = result.rfind('/', pos-1);
+		if (pos != result.npos) {
+			result.erase(pos);
+			result.append("/share/backscrub/");
+			result.append(type);
+			result.append("/");
+			result.append(provided);
```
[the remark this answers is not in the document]
phlash: Yep, as above!

app/deepseg.cc
```diff
@@ -256,6 +263,57 @@ static bool is_number(const std::string &s) {
...
+			if (std::ifstream(result).good())
+				return result;
+			// development folder?
+			result.erase(pos);
+			result.append("/");
+			result.append(type);
+			result.append("/");
+			result.append(provided);
```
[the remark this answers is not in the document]
phlash: ..and again, yep, as above.

app/deepseg.cc
```diff
@@ -256,6 +263,57 @@ static bool is_number(const std::string &s) {
 	return !s.empty() && std::all_of(s.begin(), s.end(), ::isdigit);
 }
 
+std::string resolve_path(const std::string& provided, const std::string& type) {
```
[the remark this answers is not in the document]
phlash: Looks like `std::optional` is c++17 or higher, looks like it's available in MSVC so yep, nice and clean.

app/deepseg.cc
```diff
@@ -256,6 +263,57 @@ static bool is_number(const std::string &s) {
...
+			result.erase(pos);
+			result.append("/");
+			result.append(type);
+			result.append("/");
+			result.append(provided);
+			if (std::ifstream(result).good())
+				return result;
+		}
+	}
+	// 4. XDG standard locations...maybe later?
```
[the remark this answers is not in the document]
phlash: OK, this would typically mean `~/.local/share/backscrub/...`, but with environmental overrides. Will do.

app/deepseg.cc
```diff
@@ -256,6 +263,57 @@ static bool is_number(const std::string &s) {
...
+	result.clear();
+	return result;
```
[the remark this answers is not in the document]
phlash: along with `std::optional`, yep.

app/deepseg.cc
```diff
@@ -379,7 +437,7 @@ int main(int argc, char* argv[]) try {
 	if (showUsage) {
 		fprintf(stderr, "\n");
 		fprintf(stderr, "usage:\n");
-		fprintf(stderr, "  deepseg [-?] [-d] [-p] [-c <capture>] [-v <virtual>] [-w <width>] [-h <height>]\n");
+		fprintf(stderr, "  backscrub [-?] [-d] [-p] [-c <capture>] [-v <virtual>] [-w <width>] [-h <height>]\n");
```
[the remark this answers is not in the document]
phlash: Doh - one day I'll remember there are sane ways to do this.. yep!

app/deepseg.cc
phlash: along with `std::optional` return type change above, then `s_backg` and `s_model` are also optional so this should clean up to avoid `.empty()` checks...

app/deepseg.cc
phlash: Agreed, we ought to stick with `const std::string &` unless it's an external API that uses `const char *`, will fix for background loader too.

app/deepseg.cc
BenBE: If it's not too much complexity code-wise go for it.

app/deepseg.cc
BenBE: Yes. Best split this into another change. But assuming that without a path this is meant to reside in `/dev` might be worthwhile in the meantime.

app/deepseg.cc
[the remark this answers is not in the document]
BenBE: I was more thinking about having this lib as optional dependency giving additional performance IFF the user chooses to while compiling it. If it's not configured just fall back to the old-fashioned way. Can be added later in a separate change …

app/deepseg.cc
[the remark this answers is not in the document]
BenBE: You see why I asked for C++17 a while back? :laughing:

app/deepseg.cc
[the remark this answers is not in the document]
BenBE: This would even do with normal `std::string` meaning to construct an empty string from scratch …

phlash:
OK, I think all that's covered..

app/deepseg.cc
```diff
@@ -213,11 +215,11 @@ class CalcMask final {
 	long maskns;
 	long loopns;
 
-	CalcMask(const char *modelname,
+	CalcMask(const std::string& modelname,
 			 size_t threads,
 			 size_t width,
 			 size_t height) {
-		maskctx = bs_maskgen_new(modelname,threads,width,height,nullptr,onprep,oninfer,onmask,this);
+		maskctx = bs_maskgen_new(modelname.c_str(),threads,width,height,nullptr,onprep,oninfer,onmask,this);
```
BenBE, line 222: Might be worth carrying over that type change to use STL types here too.

app/deepseg.cc
```diff
@@ -274,12 +276,21 @@ std::string resolve_path(const std::string& provided, const std::string& type) {
 	// 1. exactly what was provided
 	if (std::ifstream(provided).good())
 		return provided;
+	// to emulate PATH search behaviour (rule of least surprise), we stop here if provided has path separators
+	if (provided.find('/') != provided.npos)
+		return {};
+	// 1a. BACKSCRUB_PATH prefixes if set
```
BenBE: Having a bare `1a` feels like there's something missing. What about `1.` -> `0.` and `1a` -> `1.`?

app/deepseg.cc
```diff
@@ -485,22 +496,22 @@ int main(int argc, char* argv[]) try {
...
-		if (!s_backg.empty()) {
+		if (s_backg) {
 			printf("Warning: could not load background image, defaulting to green\n");
 		}
 	}
 	// default green screen background
 	cv::Mat bg = cv::Mat(height,width,CV_8UC3,cv::Scalar(0,255,0));
 
-	int lbfd = loopback_init(vcam,width,height,debug);
+	int lbfd = loopback_init(s_vcam.c_str(),width,height,debug);
```
BenBE, line 512: What about carrying the STL type over to `loopback_init`?

app/deepseg.cc
phlash: I thought I renumbered them.. will check!

app/deepseg.cc
phlash: Yep, will do.

app/deepseg.cc
phlash: Yep, will do.

app/deepseg.cc
phlash: Yes, they are renumbered :smile:

app/deepseg.cc
```diff
@@ -256,9 +265,66 @@ static bool is_number(const std::string &s) {
...
+	if (std::ifstream(result).good())
+		return result;
+	// 4. prefixed with compile-time install path
+	result = std::string() + _STR(INSTALL_PREFIX) + "/share/backscrub/" + type + "/" + provided;
+	if (std::ifstream(result).good())
+		return result;
+	// 5. relative to current binary location
+	// (https://stackoverflow.com/questions/933850/how-do-i-find-the-location-of-the-executable-in-c)
+	char binloc[1024];
+	ssize_t n = readlink("/proc/self/exe", binloc, sizeof(binloc));
```
BenBE, line 304: Location of procfs mount may vary. Also not supported on some platforms.

BenBE (approved):
LGTM. Mostly esoteric hints left, regarding cross-platform support …

app/deepseg.cc
phlash: Yep - best option from a poor set of choices sadly, you would think this would be a trivial thing to find out..

BenBE: There's at least some existing code in [other projects](https://github.com/htop-dev/htop/blob/68c00b9cdb8ea9a841f322effb385efce1436573/configure.ac#L448-L456) \*cough\* I can point at … ;-)

phlash: We can avoid `/proc` by using the auxiliary vector, which might be a better option: https://stackoverflow.com/a/23621200. I've been playing with this and resolving relative paths via `getcwd()` and `realpath()`, which seems pretty reliable.

BenBE: The aux vector and in particular the `AT_EXECFN` stuff seems to be glibc specials and thus likely not portable. With the `/proc` approach we have two advantages: the value can't be interfered by userland and it can be easily changed using cmake (when detecting the configuration for the system).

Issue #120, opened 2021-11-03
Test builds with later Tensorflow Lite releases


phlash:
TFLite are now up to 2.6.1 (stable) and beyond in testing releases: https://github.com/tensorflow/tensorflow/releases

These may offer fixes for newer toolchains (GCC 11+, Clang 12+), in particular on more bleeding edge distros such as Manjaro.

Ticket here is to try these later versions of TFLite against newer toolchains, and ensure `backscrub` still actually works :smile:

phlash:
Manjaro, GCC 11.1.0: TFLite v2.6.0 - builds ok :+1:
Debian (stable), GCC 10.2.0: TFLite v2.6.0 - build ok :+1: , runs ok :clap: 

PR incoming - would appreciate others testing on their distro.

BenBE:
Updated `main` and `experimental` to TFLite 2.6.0: works fine on Lubuntu 20.04 LTS for me, no obvious regressions.

Didn't test with any more recent TFLite releases.

cypMon:
> Manjaro, GCC 11.1.0: TFLite v2.6.0 - builds ok 

Probably doesn't help a lot, but builds and runs great under Arch Linux with the same versions.

BenBE:
What might be interesting is if the code still works properly with the [upcoming 2.7 release](https://github.com/tensorflow/tensorflow/releases/tag/v2.7.0-rc1).

jjsarton:
Ferora, GCC 12.1.1: TFLite v2.8.0 - builds ok

Issue #119, opened 2021-10-31
build errors: std::numeric_limits & _mm_storeu_si32


thomas725:
I have a few problems that prevent me from building your project:

```
path/backscrub/build/abseil-cpp/absl/synchronization/internal/graphcycles.cc: In member function ‘void absl::lts_2020_02_25::synchronization_internal::GraphCycles::RemoveNode(void*)’:
path/backscrub/build/abseil-cpp/absl/synchronization/internal/graphcycles.cc:451:26: error: ‘numeric_limits’ is not a member of ‘std’
  451 |   if (x->version == std::numeric_limits<uint32_t>::max()) {
      |                          ^~~~~~~~~~~~~~
path/backscrub/build/abseil-cpp/absl/synchronization/internal/graphcycles.cc:451:49: error: expected primary-expression before ‘>’ token
  451 |   if (x->version == std::numeric_limits<uint32_t>::max()) {
      |                                                 ^
path/backscrub/build/abseil-cpp/absl/synchronization/internal/graphcycles.cc:451:52: error: ‘::max’ has not been declared; did you mean ‘std::max’?
  451 |   if (x->version == std::numeric_limits<uint32_t>::max()) {
      |                                                    ^~~
      |                                                    std::max
In file included from /usr/include/c++/11.1.0/algorithm:62,
                 from path/backscrub/build/abseil-cpp/absl/synchronization/internal/graphcycles.cc:38:
/usr/include/c++/11.1.0/bits/stl_algo.h:3467:5: note: ‘std::max’ declared here
 3467 |     max(initializer_list<_Tp> __l, _Compare __comp)
      |     ^~~
In file included from path/backscrub/build/xnnpack/src/f32-vbinary/gen/vadd-minmax-sse-x4.c:15:
path/backscrub/build/xnnpack/src/xnnpack/intrinsics-polyfill.h:27:6: error: conflicting types for ‘_mm_storeu_si32’; have ‘void(const void *, __m128i)’
   27 | void _mm_storeu_si32(const void* address, __m128i v) {
      |      ^~~~~~~~~~~~~~~
In file included from /usr/lib/gcc/x86_64-pc-linux-gnu/11.1.0/include/xmmintrin.h:1316,
                 from path/backscrub/build/xnnpack/src/f32-vbinary/gen/vadd-minmax-sse-x4.c:12:
[...]

Issue #103, opened 2021-08-24
Cmake error


hellojaccc:
I tried building on arch, with #93's manual. This error occured. How can i fix this?

```
[ 58%] Linking C static library libXNNPACK.a
[ 87%] Built target XNNPACK
make: *** [Makefile:156: all] Error 2
```

floe:
I guess you'll have to build with `VERBOSE=1 make` and post the actual error before we can make any guess here.

hellojaccc:
Sorry, i think i missed one step in #93 . It now has build correctly.

I have this error now though :(
```
ERROR: Could not open 'models/segm_full_v679.tflite'.
error: unable to load model from file: 'models/segm_full_v679.tflite'.
Error: Could not create mask context
```

hellojaccc:
#101, Manually loading model works

Issue #93, opened 2021-06-25
compile with gcc-11


gled-rs:
More of a notes than an issue but at this point tensorflow does not compile with gcc 11, so if needed here are the notes on how to compile.

Changes needed:

```
mkdir build
cd build
cmake ../
```

then in:
- abseil-cpp/absl/synchronization/internal/graphcycles.h and ruy/ruy/block_map.h add at the top:
```
 #include <limits>
 #include <stdexcept>
```
- xnnpack/src/xnnpack/intrinsics-polyfill.h replace line 15 :
`#if (defined(__GNUC__) && !defined(__clang__) && !defined(__INTEL_COMPILER)) || \`
by
`#if (defined(__GNUC__) && !defined(__clang__) && !defined(__INTEL_COMPILER) && __GNUC__ < 11) || \`

BenBE:
Care to upstream those comments to the TensorFlow guys? Just in case there aren't open issues for those two changes there yet.

gled-rs:
> Care to upstream those comments to the TensorFlow guys? Just in case there aren't open issues for those two changes there yet.

they are well aware of the compile issue, that is where I found the workaround ( split in 2 different sources ).

I just thought of adding the note here to save some time for someone who would like to compile with gcc11 ( arch users for example ).

phlash:
Is this still an issue for newer Tensorflow versions? We're on 2.8.0 at present.

BenBE:
AFAIR we now got problems with GCC 12, cf. #145 …

PR #121, opened 2021-11-03
update Tensorflow to v2.6.0

phlash:
As per #120, this PR updates Tensorflow to v2.6.0, which appears to fix a number of build issues (#93, #103, #119) on distros that have more up to date compilers (GCC-11, clang-12).


BenBE (approved):

phlash:
Thanks @BenBE !

PR #122, opened 2021-11-04
bug fix: avoid OpenCV display calls on multiple threads, merge debug video windows

phlash:
These changes ensure OpenCV debug display calls (`cv::imshow()`) are on the main thread only, in an attempt to fix crashes experienced by others when using animated backgrounds. We also merge the separate background debug window into the main debug window (as picture-in-picture) and add keyboard controls to turn this on/off, along with FPS display, so users can grab the debug window as a clean image if required.

### DIFF
--- a/app/background.cc
+++ b/app/background.cc
@@ -18,8 +18,10 @@ struct background_t {
     int frm;
     double fps;
     cv::Mat raw;
+    cv::Mat thm;
     std::thread thr;
     std::mutex mux;
+    std::mutex mtm;
 };
 
 // Internal video reader thread
@@ -27,7 +29,6 @@ static void read_thread(std::shared_ptr<background_t> pbkd) {
     if (pbkd->dbg) fprintf(stderr, "background: thread start\n");
     auto last = std::chrono::steady_clock::now();
     auto proc = last;
-    double min=0.0, max=0.0;
     while (pbkd->run) {
         // read new frame - we use a temporary buffer for two reasons:
         // - we can read unlocked, thus we are not impacted by blocking backends (eg: V4L2)
@@ -42,21 +43,20 @@ static void read_thread(std::shared_ptr<background_t> pbkd) {
             }
             // grab timing point
             auto now = std::chrono::steady_clock::now();
-            // display thumbnail frame with overlay info if double debug enabled
+            // generate thumbnail frame with overlay info if double debug enabled
             if (pbkd->dbg > 1) {
                 char msg[40];
                 long nsec = std::chrono::duration_cast<std::chrono::nanoseconds>(now-last).count();
                 double fps = 1e9/(double)nsec;
-                // allow a few settling frames..
-                if (pbkd->frm > 10) {
-                    min = (0.0==min || fps<min) ? fps : min;
-                    max = (0.0==max || fps>max) ? fps : max;
+                {
+                    std::unique_lock<std::mutex> hold(pbkd->mtm);
+                    cv::resize(grab, pbkd->thm, cv::Size(160,120));
+                    snprintf(msg, sizeof(msg), "FPS:%0.1f", fps);
+                    cv::putText(pbkd->thm, msg, cv::Point(5,15), cv::FONT_HERSHEY_PLAIN, 1.0, cv::Scalar(0,255,0));
+                    snprintf(msg, sizeof(msg), "FRM:%05d", fps, pbkd->frm);
+                    cv::putText(pbkd->thm, msg, cv::Point(5,30), cv::FONT_HERSHEY_PLAIN, 1.0, cv::Scalar(0,255,0));
+                    cv::putText(pbkd->thm, "Background", cv::Point(5,pbkd->thm.rows-5), cv::FONT_HERSHEY_PLAIN, 1.0, cv::Scalar(0,255,0));
                 }
-                snprintf(msg, sizeof(msg), "FPS:%0.1f<%0.1f<%0.1f FRM:%05d", min, fps, max, pbkd->frm);
-                cv::Mat frame;
-                cv::resize(grab, frame, cv::Size(320,240));
-                cv::putText(frame, msg, cv::Point(5,frame.rows-5), cv::FONT_HERSHEY_PLAIN, 1.0, cv::Scalar(0,255,0));
-                cv::imshow("Background", frame);
             }
             last = now;
             // wait for next frame, some sources are real-time, others are not, this ensures all play in real-time.
@@ -92,6 +92,7 @@ static void drop_background(background_t *pbkd) {
         // clean up
         pbkd->cap.release();
         pbkd->raw.release();
+        pbkd->thm.release();
     } else {
         // clean up
         pbkd->raw.release();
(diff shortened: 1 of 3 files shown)


app/background.cc
```diff
@@ -18,16 +18,17 @@ struct background_t {
     int frm;
     double fps;
     cv::Mat raw;
+    cv::Mat thm;
```
BenBE, line 21: Hard disks are large enough for good identifiers (or comments explaining their purpose). ;-)

app/background.cc
```diff
@@ -18,16 +18,17 @@ struct background_t {
...
     std::thread thr;
     std::mutex mux;
+    std::mutex mtm;
```
BenBE, line 24: Same here: What's the difference between those two mutexes?

app/background.cc
```diff
@@ -168,3 +169,11 @@ int grab_background(std::shared_ptr<background_t> pbkd, int width, int height, c
...
     return frm;
 }
+
+int grab_thumbnail(std::shared_ptr<background_t> pbkd, cv::Mat &out) {
+    if (!pbkd)
+        return -1;
+    std::unique_lock<std::mutex> hold(pbkd->mtm);
+    out = pbkd->thm.clone();
+    return 0;
+}
```
BenBE, line 179: Missing LF @ EOF.

BenBE (changes requested):
Optional: Can we add similar pic-in-pic for the segmentation mask too?

app/background.cc
phlash: I did wonder why I'd done that earlier - proper names coming up!

app/background.cc
phlash: as above - I'll use proper names for variables!

app/background.cc
phlash: Ta, will fix.

phlash:
> Optional: Can we add similar pic-in-pic for the segmentation mask too?

I'm sure we could - top right?

app/background.cc
```diff
@@ -11,23 +11,24 @@
...
-    int frm;
+    int frame;
     double fps;
     cv::Mat raw;
-    std::thread thr;
-    std::mutex mux;
+    cv::Mat thumb;
+    std::thread thread;
+    std::mutex rawmux;
+    std::mutex thumbmux;
```
BenBE, line 24:
```suggestion
    std::mutex rawmux;
    cv::Mat thumb;
    std::mutex thumbmux;
    std::thread thread;
```

Reordering to have all mutexes and cv::Mat before the thread object (also determines initialization order).

Maybe rename `thread` to `segmentation_thread`?

app/background.cc
```diff
@@ -36,62 +37,62 @@ static void read_thread(std::shared_ptr<background_t> pbkd) {
...
-                    min = (0.0==min || fps<min) ? fps : min;
-                    max = (0.0==max || fps>max) ? fps : max;
+                {
+                    std::unique_lock<std::mutex> hold(pbkd->thumbmux);
+                    cv::resize(grab, pbkd->thumb, cv::Size(160,120));
+                    snprintf(msg, sizeof(msg), "FPS:%0.1f", fps);
+                    cv::putText(pbkd->thumb, msg, cv::Point(5,15), cv::FONT_HERSHEY_PLAIN, 1.0, cv::Scalar(0,255,255));
+                    snprintf(msg, sizeof(msg), "FRM:%05d", fps, pbkd->frame);
+                    cv::putText(pbkd->thumb, msg, cv::Point(5,30), cv::FONT_HERSHEY_PLAIN, 1.0, cv::Scalar(0,255,255));
+                    cv::putText(pbkd->thumb, "Background", cv::Point(5,pbkd->thumb.rows-5), cv::FONT_HERSHEY_PLAIN, 1.0, cv::Scalar(0,255,255));
```
BenBE, line 58: Minimize work done under active lock. Use thread-local buffer to prepare and only lock once the thumbnail is ready.

Second thing: the thumbnail should keep the ratio from the input/output frame; maybe 1:8?

app/deepseg.cc
```diff
@@ -630,9 +634,53 @@ int main(int argc, char* argv[]) try {
...
+			char status[80];
+			snprintf(status, sizeof(status), "MainFPS: %5.2f AiFPS: %5.2f", mfps, afps);
+			cv::putText(test, status, cv::Point(5,test.rows-5), cv::FONT_HERSHEY_PLAIN, 1.0, cv::Scalar(0,255,255));
+		}
+		// keyboard help
+		if (showHelp) {
+			static const char *help[] = {
+				"Keyboard help:",
+				" q: Quit",
+				" s: Switch filter on/off",
```
BenBE, line 648: Consistent upper/lower case …

BenBE (changes requested):
Please have a look to have consistent space after comma for arguments and similar style issues.
Functionality itself looks mostly good already, except for some performance/locking issues that keep the lock for too long and thus may degrade performance.

app/background.cc
phlash: will do.

app/background.cc
phlash: good point, copying `cv::Mat` objects is low cost (header only), so working in a local and copying out would be fine.

app/deepseg.cc
phlash: I was trying to indicate why the letters are chosen, not sure this helps :question:

phlash:
> Please have a look to have consistent space after comma for arguments and similar style issues.

If you don't mind, I'll do that in a separate PR, which covers the wider codebase?

BenBE:
> > Please have a look to have consistent space after comma for arguments and similar style issues.
> 
> If you don't mind, I'll do that in a separate PR, which covers the wider codebase?

Sure, not complains with splitting this into a separate PR.

BenBE (approved):

phlash:
Over in https://github.com/floe/backscrub/pull/124#discussion_r744138574, referring to https://github.com/floe/backscrub/blob/44f94f3820c22b4b53f3a1599468867bb84249ef/app/background.cc#L28
> This somewhat counter-acts the reference counting done by the std::shared_ptr, thus extra care should be taken, that there is always one live reference of the original std::shared_ptr while any code of the read_thread function executes; otherwise risking UB. If you really have to pass a plain pointer here, make that point to the std::shared_ptr and copy the actual std::shared_ptr inside the routine to ensure this required local copy to ensure correct object lifetimes.

The intended behaviour was to terminate the `read_thread` and clean up memory when the last reference to the background object is dropped (presently at the end on `main`). To this end, the object is created with a _deleter function_ that performs this task, but it wasn't being called, resulting in random crashes on exit, as the thread itself held an additional reference. This change was to prevent this erroneous behaviour. There is no risk of UB as the deleter function will always be called before the object is freed.

BenBE:
If that's the case, we're probably better of (even if this takes some extra checks) to use a `std::weak_ptr` inside the thread, that needs to be grabbed explicitly before use. That way no owning reference is retained inside the thread (unless needed) and the main thread can properly delete the object upon termination.

Cf. https://en.cppreference.com/w/cpp/memory/weak_ptr/lock for an usage example.

Additionally all threads should be joined. In combination with releasing the central `std::shared_ptr` this might signal another condition for thread termination.

phlash:
OK, this is done - minor change so checked in without a PR (also I forgot which repo was upstream of my local work area!)

PR #124, opened 2021-11-06
Syntax fixes

phlash:
Code tidy up, starting with white space in expressions..


app/deepseg.cc
```diff
@@ -328,64 +328,64 @@ int main(int argc, char* argv[]) try {
...
+	bool showHelp        = false;
+	size_t threads       = 2;
+	size_t width         = 640;
+	size_t height        = 480;
+	const char *back     = nullptr;
+	const char *vcam     = "/dev/video1";
+	const char *ccam     = "/dev/video0";
+	bool flipHorizontal  = false;
+	bool flipVertical    = false;
+	int fourcc           = 0;
```
BenBE, line 346: Not quite a fan of those tabulated indentations.

app/background.cc
```diff
@@ -25,7 +25,7 @@ struct background_t {
 };
 
 // Internal video reader thread
-static void read_thread(std::shared_ptr<background_t> pbkd) {
+static void read_thread(background_t *pbkd) {
```
BenBE: This somewhat counter-acts the reference counting done by the `std::shared_ptr`, thus extra care should be taken, that there is always one live reference of the original `std::shared_ptr` while any code of the `read_thread` function executes; otherwise risking UB. If you really have to pass a plain pointer here, make that point to the `std::shared_ptr` and copy the actual `std::shared_ptr` inside the routine to ensure this required local copy to ensure correct object lifetimes.

app/deepseg.cc
phlash: eww, those were supposed to be spaces..

app/background.cc
phlash: Err, that's not part of this PR? It's in #122.. happy to discuss over there :smile:

app/deepseg.cc
BenBE: Yes. I'd prefer just `var = value;` with single spaces. Having the values all aligned at one column forces changes too all lines if any of those lines had its alignment changed.

app/background.cc
BenBE: Saw too late. But will do.

BenBE (approved):

PR #126, opened 2021-11-20
Use correct aspect ratio and allow for wide models

### LINKED ISSUE #118: Calculation of model ratio seems to be wrong.
As far as I understand, ```ctx.input.cols``` is the width of the model input, ```ctx.input.rows``` its height.
The ratio is set as
https://github.com/floe/backscrub/blob/36083432c723eb0bc3f2d427fa9335a152a33111/lib/libbackscrub.cc#L226
Then the ROI width is calculated as ```height/ctx.ratio```.
https://github.com/floe/backscrub/blob/36083432c723eb0bc3f2d427fa9335a152a33111/lib/libbackscrub.cc#L229

```height/ctx.ratio``` results in ```height/ctx.input.cols/ctx.input.rows```.
This seems wrong to me, the ratio needs be be inverted to
```ctx.ratio = ctx.input.rows/ctx.input.cols```

That is the result you get when equating the ratios and solving for the width:
(assuming number of cols is a width, number of rows is a height)
```
   width / height = ctx.input.cols / ctx.input.rows
=> width = (ctx.input.cols/ctx.input.rows) * height
=> width = height * (input.cols / input.rows)
=> width = height / (1 / (input.cols / input.rows) )
=> width = height / (input.rows / input.cols )
```

The inverted ratio seems to be correct for all places where the ratio is used.

That fix helps with the landscape version of the mlkit / mediapipe model (published under Apache 2.0):
https://google.github.io/mediapipe/solutions/models.html#selfie-segmentation

By the way, the meet model seems to be a landscape model as well according to the input tensor metadata contrary to the information on the model card?

Edit: Only changing the ratio will cause a crash with a camera that is only providing a 4:3 image (e.g. 640x480) and no 16:9 image like 1280x720.

progandy:
The aspect ratio calculation was reversed which resultet in suboptimal
model area use for landscape models.
Fixing that also requires additional handling of models that are wider
than the input frame.

Fixes #118

For now I chose to always use the full model height:
For wide frames and narrow models, only the center of the frame is used (the previous behavior). 
For narrow frames and wide models or the frame aspect being the same as the model, the complete frame is fit into the model.

### DIFF
--- a/lib/libbackscrub.cc
+++ b/lib/libbackscrub.cc
@@ -46,7 +46,10 @@ struct backscrub_ctx_t {
 	cv::Mat mroi;
 	cv::Mat ofinal;
 	cv::Size blur;
+	cv::Mat in_u8_bgr;
+	cv::Rect in_roidim;
 	float ratio;
+	float frameratio;
 };
 
 // Debug helper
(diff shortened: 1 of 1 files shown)


BenBE (approved):
LGTM.

BenBE:
I think with two reviews approving this change we're gtg …

floe:
No need to wait for me, still hopelessly under water ATM... :-/

PR #127, opened 2021-12-07
Add convenience script for (re-)building

theCalcaholic:
I've created this script for myself, but figured, others might benefit as well.

Tested on Linux (Ubuntu 20.04, but that shouldn't matter)

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,0 +1,73 @@
+#!/usr/bin/env bash
+
+(
+
+if [[ " $* " =~ " "("-h"|"--help")" " ]]
+then
+  echo "Update and (re-)build backscrub"
+  echo "USAGE:"
+  echo "  ./build.sh [OPTIONS]"
+  echo ""
+  echo "  Options:"
+  echo "    --h, --help  Show this help message and exit"
+  echo "    --no-update  Skip fetching updates from github"
+  exit 0
+fi
+
+# Get the root directory of the repository
+wd="$(cd "$(dirname $0)" || exit 1; pwd)"
+
+echo "working directory: $wd"
+cd "$wd"
+
+# Make sure we have the latest version of backscrub
+echo "Fetching latest version..."
+[[ " $* " =~ " --no-update" ]] || git pull --recurse-submodules
+echo "..done"
+echo ""
+
+
+# Backup the current build directory (if present)
+if [[ -d "./build" ]]
+then
+  echo "Backing up build directory..."
+  [[ ! -d "./build_bk" ]] || rm -rf "./build_bk"
+  mv ./build ./build_bk
+  echo "..done"
+  echo ""
+fi
+
+# Restore the backup on error
+trap 'rm -rf ./build && mv ./build_bk ./build' SIGHUP SIGINT SIGTERM
+
+# Abort on errors
+set -e
+
+echo "Building backscrub..."
+
+#mkdir build
+cd build
+
+# Build backscrub
+cmake ..
+make -j "$(nproc || echo 4)"
+ln -s ../models models
+
+echo "..done"
+echo ""
+
+echo "Cleaning up..."
+# Delete the backup on success
+[[ ! -d "./build_bk" ]] || rm -rf "./build_bk"
+echo "..done"
+echo ""
+
+echo "All done. You can run backscrub like this:"
+echo "    cd \"$wd/build\" && ./backscrub -h"
+echo ""
+echo "Or you can add the following line to your ~/.bashrc, to get a backscrub command:"
+echo "backscrub() { pushd \"$wd/build\" > /dev/null && { ./backscrub \"\$@\"; popd > /dev/null; }; }"
+
+
+)
+


build.sh
```diff
@@ -0,0 +1,73 @@
+#!/usr/bin/env bash
+
+(
```
BenBE, line 3: Why this extra-long block spawning the whole file?

build.sh
```diff
@@ -0,0 +1,73 @@
...
+  mv ./build ./build_bk
+  echo "..done"
+  echo ""
+fi
+
+# Restore the backup on error
+trap 'rm -rf ./build && mv ./build_bk ./build' SIGHUP SIGINT SIGTERM
+
+# Abort on errors
+set -e
```
BenBE, line 44: Shouldn't this be at the start of the script?

build.sh
```diff
@@ -0,0 +1,73 @@
...
+  echo "  ./build.sh [OPTIONS]"
+  echo ""
+  echo "  Options:"
+  echo "    --h, --help  Show this help message and exit"
+  echo "    --no-update  Skip fetching updates from github"
+  exit 0
+fi
+
+# Get the root directory of the repository
+wd="$(cd "$(dirname $0)" || exit 1; pwd)"
```
BenBE, line 18:
```suggestion
wd="$(cd "$(dirname "$0")" || exit 1; pwd)"
```

Though I'm not sure how calling this script as `--foo/foo.sh` would react with the above command …

build.sh
```diff
@@ -0,0 +1,73 @@
...
+then
+  echo "Backing up build directory..."
+  [[ ! -d "./build_bk" ]] || rm -rf "./build_bk"
+  mv ./build ./build_bk
+  echo "..done"
+  echo ""
+fi
+
+# Restore the backup on error
+trap 'rm -rf ./build && mv ./build_bk ./build' SIGHUP SIGINT SIGTERM
```
BenBE, line 41: Only remove temporary files when you actually created them. If there's some `build_bk` directory this ensures you're losing data …

I'd prefer to have the script just act on the `build` directory as that's the only directory that should be relevant for building.

build.sh
```diff
@@ -0,0 +1,73 @@
...
+
+# Abort on errors
+set -e
+
+echo "Building backscrub..."
+
+#mkdir build
+cd build
```
BenBE, line 49: Should be relative to the script path …

build.sh
```diff
@@ -0,0 +1,73 @@
...
+
+# Restore the backup on error
+trap 'rm -rf ./build && mv ./build_bk ./build' SIGHUP SIGINT SIGTERM
+
+# Abort on errors
+set -e
+
+echo "Building backscrub..."
+
+#mkdir build
```
BenBE, line 48:
```suggestion
mkdir -p build
```

Also make this relative to the script path …

build.sh
```diff
@@ -0,0 +1,73 @@
...
+cmake ..
+make -j "$(nproc || echo 4)"
+ln -s ../models models
+
+echo "..done"
+echo ""
+
+echo "Cleaning up..."
+# Delete the backup on success
+[[ ! -d "./build_bk" ]] || rm -rf "./build_bk"
```
BenBE, line 61: Only remove this if you initially created it …

build.sh
```diff
@@ -0,0 +1,73 @@
...
+echo "..done"
+echo ""
+
+echo "All done. You can run backscrub like this:"
+echo "    cd \"$wd/build\" && ./backscrub -h"
+echo ""
+echo "Or you can add the following line to your ~/.bashrc, to get a backscrub command:"
+echo "backscrub() { pushd \"$wd/build\" > /dev/null && { ./backscrub \"\$@\"; popd > /dev/null; }; }"
```
BenBE, line 69:
```suggestion
echo "backscrub() { ( cd \"$wd/build\" && ./backscrub \"\$@\" ) }"
```

BenBE (changes requested):

build.sh
theCalcaholic: That's so the `cd` commands don't affect the calling context (otherwise you might end up in a different directory than your pwd before starting the script)

BenBE: Unless the caller sources this script it's executed as a sub-process anyway, thus no need for starting yet another sub-shell here IMHO.

theCalcaholic (commented):
I have attempted to address all issues you mentioned now :)

theCalcaholic:
@BenBE Could you explain what is unclear about the usecase (as you added the "needs usecase" label)?

The idea is to make the setup of backscrub less error prone and more convenient, especially for less technical users.

theCalcaholic (commented):
I removed the subshell

phlash:
Howdy, thanks for this :smile:  Given the intention is to make it easier for non-technical folks, would it be useful to ensure dependencies are installed as well as automating the instructions in the README? If so, perhaps using a CMake based mechanism similar to (if not exactly the same!) as Tensorflow uses would be effective?

Can I assume that backing up the `build` folder is to avoid lengthy downloads/rebuilds of Tensorflow? If so, it might be better to leverage the support for out-of-tree Tensorflow we have already in CMakeFiles.list?

Please note: just questions for later, I'm happy this is merged as-is, then we can think about my random questions!

BenBE:
> @BenBE Could you explain what is unclear about the usecase (as you added the "needs usecase" label)?

I added the label for the other devs to have a look at this and chime in pro/con this change.

Given the comments from @phlash I think resolving the dependency re-download issue should be addressed; apart from that there's only minor stylistic things (missing escapes for the path argument in the trap call, e.g.) that I'd to in a later review.

theCalcaholic:
@phlash Thanks, these are very good points :)

> Given the intention is to make it easier for non-technical folks, would it be useful to ensure dependencies are installed as well as automating the instructions in the README? If so, perhaps using a CMake based mechanism similar to (if not exactly the same!) as Tensorflow uses would be effective?

That's a good idea. I'm not quite sure how easy it would be to do this distro-agnostic, though (and my knowledge about CMake is quite limited), so we'd need to see how to implement this.

> Can I assume that backing up the `build` folder is to avoid lengthy downloads/rebuilds of Tensorflow? If so, it might be better to leverage the support for out-of-tree Tensorflow we have already in CMakeFiles.list?

No, I added that just to avoid breaking the setup. Everything is rebuilt anyways. However, if you use the script for updating backscrub and something breaks (e.g. a dependency is missing or an environment variable is set wrongly), users will still have their working pre-update version.

PR #135, opened 2022-01-24
Multi-pass processing

BenBE:
This branch will introduce multi-pass processing of the input image for cases when the internal face detection model cannot handle the full input image in one go due to the aspect ratio of the model compared to the aspect ratio of the input image.

This PR also contains a major cleanup of the code base by making the code style more consistent and splitting things into multiple files for more digestible chunks of code.


app/deepseg.cc
```diff
@@ -95,6 +96,8 @@ int main(int argc, char* argv[]) try {
 			flipHorizontal = !flipHorizontal;
 		} else if (args[arg] == "-V") {
 			flipVertical = !flipVertical;
+		} else if (args[arg] == "-M") {
+			multipass = !multipass;
```
phlash, line 100: This toggles the multipass flag if `-M` is used multiple times.. intended or confusing?

phlash:
Do we want to consider parallel processing the multiple passes? IMO this is probably not worth the extra pain unless we run very low on time budget..

app/deepseg.cc
BenBE: That's the same as with the other existing flags …

BenBE:
> Do we want to consider parallel processing the multiple passes? IMO this is probably not worth the extra pain unless we run very low on time budget..

I'll do this as one process in the first version. We can still upgrade to multi-threading this processing in a second update.

Issue #138, opened 2022-01-24
Optimization potential


BenBE:
Doing some rudimentary profiling I noticed, we are wasting a huge amount of time of the main thread in essentially 3 functions:

1. `alpha_blend` (~40%)
2. `convert_rgb_to_yuyv` (~25%)
3. `cap.retrieve` (~20%)

Values are relative time for `main`, which takes up roughly the same amount of time, we actually spend processing images (`main` ~32.75%, `bs_maskgen_process` ~33.95%, total runtime).

On the positive: `bs_maskgen_process` spends ~95% of the time waiting for processing by TFLite.

FWIW: Timings heavily affected by running under callgrind, but looking at the code of `alpha_blend` and `convert_rgb_to_yuyv` I'm not really surprised of these results. A rewrite of these functions using some vectoring should yield quite a bit of improvement.

floe:
Ah, that does sound like SSE and friends could do a good job speeding things up. I guess the most portable way to do that is using gcc intrinsics?

BenBE:
Either GCC intrinsics (as portable as they are) or we try to look into things and see if OpenCV/XNN already offers those two methods with an optimized version we could use.

phlash:
I'm aware that those were written because OpenCV doesn't do them.. but happy to be proven wrong if we up the minimum OCV version :smile: 

Found this Q&A: https://answers.opencv.org/question/211941/how-to-use-simd-feature-of-opencv-as-third-party-library/

I'm also aware that OCV will try to use OpenCL and GPU if they are available (it collided with TFLite when used in multiple threads!).. more research needed!

There is also [libvolk](https://www.libvolk.org) for portable SIMD kernels (as used by Gnuradio and others), which is a standard package on many Linuxen.

BenBE:
I don't have any objections on using a library for those particular vectorized kernels, as doing those optimizations ourselves would be more work than sensible.

Also check if other places in the current code base (e.g. the conv_bias stuff) could benefit from using the library.

martok:
Just talked with @BenBE, time for my regularly scheduled one-off maybe-useful comment :)

>    `alpha_blend` (~40%)

OpenCV has a well-hidden function for that, [blendLinear](https://docs.opencv.org/4.5.5/d7/d1b/group__imgproc__misc.html#ga5e76540a679333d7c6cd0617c452c23d). Only difficulty here would be that it requires the mask and anti-mask, both as float.
[...]

PR #139, opened 2022-01-30
Performance enhancements

phlash:
Specifically to `alpha_blend` and `convert_rgb_to_yuyv` as per #138 

This first draft addresses `alpha_blend`, reducing run time by a factor of 3-4x simply by separating the function and turning on full compiler optimisation __for the machine that we are compiling on__ (`-march=native`). This may be a problem if compiling for a different target that doesn't have the same CPU features...

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -58,8 +58,8 @@ add_subdirectory(${TENSORFLOW}/tensorflow/lite
 
 # build backscrub code
 add_compile_definitions(DEEPSEG_VERSION=${DEEPSEG_VERSION} INSTALL_PREFIX=${CMAKE_INSTALL_PREFIX})
+add_compile_options(-march=native -O9)
 include_directories(BEFORE .)
-set(CMAKE_CXX_STANDARD 17)
 
 add_library(backscrub
   lib/libbackscrub.cc
@@ -87,8 +87,6 @@ add_executable(deepseg
   app/calcmask.cc
   app/utils.cc
 )
-# special compile options for blend.cc to use SIMD optimiser
-set_source_files_properties(app/blend.cc PROPERTIES COMPILE_OPTIONS "-march=native;-O9")
 
 set_target_properties(deepseg PROPERTIES OUTPUT_NAME backscrub)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -83,9 +83,12 @@ target_link_libraries(videoio)
 add_executable(deepseg
   app/deepseg.cc
   app/background.cc
+  app/blend.cc
   app/calcmask.cc
   app/utils.cc
 )
+# special compile options for blend.cc to use SIMD optimiser
+set_source_files_properties(app/blend.cc PROPERTIES COMPILE_OPTIONS "-march=native;-O9")
 
 set_target_properties(deepseg PROPERTIES OUTPUT_NAME backscrub)
 
--- a/app/blend.cc
+++ b/app/blend.cc
@@ -1,0 +1,35 @@
+#include "utils.h"
+
+cv::Mat alpha_blend(const cv::Mat& srca, const cv::Mat& srcb, const cv::Mat& mask) {
+	// alpha blend two (8UC3) source images using a mask (8UC1, 255=>srca, 0=>srcb), adapted from:
+	// https://www.learnopencv.com/alpha-blending-using-opencv-cpp-python/
+	// "trust no-one" => we're about to mess with data pointers
+	assert(srca.rows == srcb.rows);
+	assert(srca.cols == srcb.cols);
+	assert(mask.rows == srca.rows);
+	assert(mask.cols == srca.cols);
+
+	assert(srca.type() == CV_8UC3);
+	assert(srcb.type() == CV_8UC3);
+	assert(mask.type() == CV_8UC1);
+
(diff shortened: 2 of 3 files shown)


BenBE:
Could you please rebase onto experimental?

phlash:
> Could you please rebase onto experimental?

Odd, I thought this was based on current experimental, it looks like that to me?

app/blend.cc
```diff
@@ -0,0 +1,34 @@
...
+
+	for (size_t pix = 0, npix = srca.rows * srca.cols; pix < npix; ++pix) {
+		// blending weights
+		int aw = (int)(*mptr++);
+		int bw = 255 - aw;
+
+		// blend each channel byte
+		*optr++ = (uint8_t)(( (int)(*aptr++) * aw + (int)(*bptr++) * bw ) >> 8);
+		*optr++ = (uint8_t)(( (int)(*aptr++) * aw + (int)(*bptr++) * bw ) >> 8);
+		*optr++ = (uint8_t)(( (int)(*aptr++) * aw + (int)(*bptr++) * bw ) >> 8);
```
BenBE, line 31: Correctness issue: `255(a)*255(aw)+0(bw)*0(bw) == 0xFE01` -> `0xFF != 0xFE`.

If you scale this by 257 and right-shift by 16 instead of 8 you're fine.

```suggestion
		*optr++ = (uint8_t)((257*( (int)(*aptr++) * aw + (int)(*bptr++) * bw )) >> 16);
		*optr++ = (uint8_t)((257*( (int)(*aptr++) * aw + (int)(*bptr++) * bw )) >> 16);
		*optr++ = (uint8_t)((257*( (int)(*aptr++) * aw + (int)(*bptr++) * bw )) >> 16);
```

BenBE (changes requested):

app/blend.cc
phlash: This is true - I replaced the divide (by 255) with a shift to see if it was faster.. I'll go back and try that again, plus this suggestion.

BenBE:
Any updates here?

app/blend.cc
phlash: empirical results, mask step duration (msecs):
 * `>> 8` => 0.7-1.5 (as before, not accurate)
 * `/ 255` => 2.3-5.0 (gcc unable to auto-vectorize)
 * `pre-multiply aw/bw by 257`, `>> 16` => 1.4-3.8 (gcc can auto-vectorize, accurate)

Looks like the last option is the one to use.

CMakeLists.txt
```diff
@@ -85,7 +85,10 @@ add_executable(deepseg
   app/background.cc
   app/calcmask.cc
   app/utils.cc
+  app/blend.cc
```
BenBE, line 88: What about keeping this list sorted by name?
(Fine with the exception for the main function).

app/blend.cc
```diff
@@ -0,0 +1,63 @@
...
+
+	assert(srca.type() == CV_8UC3);
+	assert(srcb.type() == CV_8UC3);
+	assert(mask.type() == CV_8UC1);
+
+	const uint8_t *aptr = (const uint8_t*)srca.data;
+	const uint8_t *bptr = (const uint8_t*)srcb.data;
+	const uint8_t *mptr = (const uint8_t*)mask.data;
+
+	cv::Mat out = cv::Mat::zeros(srca.size(), srca.type());
```
BenBE, line 21: As we're writing to all elements later anyway, is there a way to just set dimensions and type, without "initializing" this? No need to change this, just an idea …

BenBE (approved):
LGTM. Just some minor ideas, nothing dramatic …

CMakeLists.txt
phlash: of course..

app/blend.cc
phlash: Can likely switch to the `Mat::create` method to avoid writing to elements.

phlash: Nice. Reduced timings to 0.6-1.8 msecs :smile:

CMakeLists.txt
```diff
@@ -83,9 +83,12 @@ target_link_libraries(videoio)
 add_executable(deepseg
   app/deepseg.cc
   app/background.cc
+  app/blend.cc
   app/calcmask.cc
   app/utils.cc
 )
+# special compile options for blend.cc to use SIMD optimiser
+set_source_files_properties(app/blend.cc PROPERTIES COMPILE_OPTIONS "-march=native;-O9")
```
BenBE, line 91: Why not for everything?

BenBE (approved):

CMakeLists.txt
phlash: I'm not sure.. I guess that if one module is targetted / optimised for the current CPU, then it may as well all be. Initially I only wanted to mess with / break one specific part of the code. I was also thinking about compiling run-time swappable implementations for different targets (this gets talked about a bit in the GCC manuals when using intrinsics) - that is probably over-engineering.

BenBE: I think as we are compiling all 3rd party deps ourselves we may as well compile everything for -march=native. Let's split this into another PR, so we could merge this one for now.

Having run-time swappable implementations is probably overkill for us right now. Let's handle this separately.

BenBE:
Please try to always rebase on the target branch instead of merging things …

Issue #140, opened 2022-02-23
Build issue on ubuntu 21.10


doktorjw:
I've just cloned the repository (used to use the deepbacksub one). On my prior version, I was able to build fine, but now ... following the direction, I get the following error:

```
[ 96%] Building CXX object tensorflow-lite/CMakeFiles/tensorflow-lite.dir/simple_memory_arena_debug_dump.cc.o
/home/dr_j/projects/backscrub/tensorflow/tensorflow/lite/simple_memory_arena_debug_dump.cc:27:43: error: ‘numeric_limits’ is not a member of ‘std’
   27 | constexpr int32_t kNodeNotAssigned = std::numeric_limits<int32_t>::max();
      |                                           ^~~~~~~~~~~~~~
/home/dr_j/projects/backscrub/tensorflow/tensorflow/lite/simple_memory_arena_debug_dump.cc:27:65: error: expected primary-expression before ‘>’ token
   27 | constexpr int32_t kNodeNotAssigned = std::numeric_limits<int32_t>::max();
      |                                                                 ^
/home/dr_j/projects/backscrub/tensorflow/tensorflow/lite/simple_memory_arena_debug_dump.cc:27:68: error: ‘::max’ has not been declared; did you mean ‘std::max’?
   27 | constexpr int32_t kNodeNotAssigned = std::numeric_limits<int32_t>::max();
      |                                                                    ^~~
      |                                                                    std::max
In file included from /usr/include/c++/11/algorithm:62,
                 from /home/dr_j/projects/backscrub/tensorflow/tensorflow/lite/simple_memory_arena_debug_dump.cc:15:
/usr/include/c++/11/bits/stl_algo.h:3467:5: note: ‘std::max’ declared here
 3467 |     max(initializer_list<_Tp> __l, _Compare __comp)
      |     ^~~
make[2]: *** [tensorflow-lite/CMakeFiles/tensorflow-lite.dir/build.make:2357: tensorflow-lite/CMakeFiles/tensorflow-lite.dir/simple_memory_arena_debug_dump.cc.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:1436: tensorflow-lite/CMakeFiles/tensorflow-lite.dir/all] Error 2
make: *** [Makefile:171: all] Error 2
```

Thoughts? Ideas?

Thx ---- john

BenBE:
This looks like a build issue with the TensorFlow Lite version we're using.

[...]

PR #141, opened 2022-02-24
update TFLite to v2.8.0

phlash:
As identified in #140. We see a fail-to-build in Tensorflow v2.7.0 on Ubuntu 21.10, fixed by updating to v2.8.0.

Please check this builds & runs on your own local environment and confirm or deny if an update to v2.8.0 is ok.
 * @phlash - works for me on Debian stable (bullseye)


BenBE:
Build works for me like a charm on Ubuntu 20.04. No obvious performance or visual regression.

PR #142, opened 2022-03-01
Display Tensorflow version during build and at runtime

phlash:
To help with diagnosis of build and runtime issues - this shows the version of Tensorflow being used in a build (after any forced checkout), and the linked runtime version.

### DIFF
--- a/app/deepseg.cc
+++ b/app/deepseg.cc
@@ -324,10 +324,9 @@ std::optional<std::string> resolve_path(const std::string& provided, const std::
 
 int main(int argc, char* argv[]) try {
 
-	printf("%s version %s\n", argv[0], _STR(DEEPSEG_VERSION));
+	printf("%s version %s (Tensorflow build: %s, run-time: %s)\n", argv[0], _STR(DEEPSEG_VERSION), _STR(TF_VERSION), bs_tensorflow_version());
 	printf("(c) 2021 by floe@butterbrot.org & contributors\n");
 	printf("https://github.com/floe/backscrub\n");
-	printf("(running with Tensorflow Lite version: %s)\n", bs_tensorflow_version());
 	timinginfo_t ti;
 	ti.bootns = timestamp();
 	int debug = 0;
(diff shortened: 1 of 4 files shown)


app/deepseg.cc
```diff
@@ -327,6 +327,7 @@ int main(int argc, char* argv[]) try {
 	printf("%s version %s\n", argv[0], _STR(DEEPSEG_VERSION));
 	printf("(c) 2021 by floe@butterbrot.org & contributors\n");
 	printf("https://github.com/floe/backscrub\n");
+	printf("(running with Tensorflow Lite version: %s)\n", bs_tensorflow_version());
```
BenBE, line 330: Should display both runtime and build time versions for TFLite.

app/deepseg.cc
```diff
@@ -327,6 +327,7 @@ int main(int argc, char* argv[]) try {
 	printf("%s version %s\n", argv[0], _STR(DEEPSEG_VERSION));
 	printf("(c) 2021 by floe@butterbrot.org & contributors\n");
```
BenBE, line 328: May update this line.

phlash: Yep -that would be tidier :smile:

app/deepseg.cc
phlash: will do.

app/deepseg.cc
```diff
@@ -324,10 +324,9 @@ std::optional<std::string> resolve_path(const std::string& provided, const std::
 
 int main(int argc, char* argv[]) try {
 
-	printf("%s version %s\n", argv[0], _STR(DEEPSEG_VERSION));
+	printf("%s version %s (Tensorflow build: %s, run-time: %s)\n", argv[0], _STR(DEEPSEG_VERSION), _STR(TF_VERSION), bs_tensorflow_version());
```
BenBE, line 327:
```suggestion
	printf("%s version %s (Tensorflow: build %s, run-time %s)\n", argv[0], _STR(DEEPSEG_VERSION), _STR(TF_VERSION), bs_tensorflow_version());
```

Issue #145, opened 2022-07-15
Compilation error on openSUSE Tumbleweed


oshikore:
backscrub used to build just fine on my system, and it does not any more. Since Tumbleweed is tumbling ahead, I can't really go back to the working state.

This is the end of the compilation, the full  output is attached.

```
[ 11%] Building CXX object _deps/flatbuffers-build/CMakeFiles/flatbuffers.dir/src/idl_parser.cpp.o
In file included from /home/oscar/Apps/teams-cam/backscrub/build/flatbuffers/include/flatbuffers/base.h:217,
                 from /home/oscar/Apps/teams-cam/backscrub/build/flatbuffers/include/flatbuffers/idl.h:24,
                 from /home/oscar/Apps/teams-cam/backscrub/build/flatbuffers/src/idl_parser.cpp:23:
/usr/include/absl/strings/string_view.h:52:26: error: ‘string_view’ in namespace ‘std’ does not name a type
   52 | using string_view = std::string_view;
      |                          ^~~~~~~~~~~
/usr/include/absl/strings/string_view.h:52:21: note: ‘std::string_view’ is only available from C++17 onwards
   52 | using string_view = std::string_view;
      |                     ^~~
/usr/include/absl/strings/string_view.h:693:8: error: ‘string_view’ does not name a type
  693 | inline string_view ClippedSubstr(string_view s, size_t pos,
      |        ^~~~~~~~~~~
/usr/include/absl/strings/string_view.h:704:11: error: ‘string_view’ does not name a type
  704 | constexpr string_view NullSafeStringView(const char* p) {
      |           ^~~~~~~~~~~
/home/oscar/Apps/teams-cam/backscrub/build/flatbuffers/include/flatbuffers/base.h:219:23: error: ‘string_view’ in namespace ‘absl’ does not name a type
  219 |         typedef absl::string_view string_view;
      |                       ^~~~~~~~~~~
In file included from /home/oscar/Apps/teams-cam/backscrub/build/flatbuffers/include/flatbuffers/idl.h:25:
/home/oscar/Apps/teams-cam/backscrub/build/flatbuffers/include/flatbuffers/flatbuffers.h:561:16: error: ‘string_view’ in namespace ‘flatbuffers’ does not name a type
  561 |   flatbuffers::string_view string_view() const {
      |                ^~~~~~~~~~~
[...]

PR #147, opened 2022-08-16
Possible fix for #145, compiling on openSUSE Tumbleweed

phlash:
This change forces an early fetch of `flatbuffers` at a newer version than Tensorflow is configured with, in the hope that it will compile cleanly on openSUSE and still work within Tensorflow..


phlash:
Closing as not useful for now (Tensorflow needs to fix their dependencies before we can do anything).

Issue #146, opened 2022-08-07
Initialization failure for strange image resolutions (like 1280x480)


jjsarton:
If backscrub is only launched with one of the -w or -h argument, backscrub crash.
Please check that both parameter width and height are stated.

BenBE:
Tracing the issue inside GDB points to some OpenCV internals:

```
terminate called after throwing an instance of 'cv::Exception'
  what():  OpenCV(4.5.4) ./modules/core/src/matrix.cpp:810: error: (-215:Assertion failed) 0 <= roi.x && 0 <= roi.width && roi.x + roi.width <= m.cols && 0 <= roi.y && 0 <= roi.height && roi.y + roi.height <= m.rows in function 'Mat'
```

The last operation attempted is in `calcmask.cc` on line 96 (branch experimental):
```c
		if(!bs_maskgen_process(maskctx, *frame_current, *mask_current)) {
```

Will need some further digging.

But initial investigation suggests that the fallback resolution used as default (`-w 640 -h 480`). Thus when using just `-w 1280` the attempted configuration is `-w 1280 -h 480`.

Attempting this resolution directly (with both parameters given) also fails for me.

BenBE:
Further debugging hints at a problem with the aspect ratio of the target video stream. 320x480 doesn't work, 640x480 works, 848x480 works, but 856x480 does not … Neither do 400x480, 560x480, 632x480 work.

The crash is independent of the background image used.

jjsarton:
I have performed some tests, and if I state the width and height correctly according to the sizes returned by v4l2-ctl --list-formats.ext  backscrub work as expected. If you check if any one the the -w or -h parameter is passed to backscrub and not both are given, you can issue an error message and exit.
It is also possible to check, at startup the camera capabilities and choose the proper height or width according to the width or height argument.

jjsarton:
I have modified a little bit app/deepseg.c

After the "size_t blur_strength = 0;"  declaration within the main() I have added "int hw = 0;".
after:
```
        			if (!width) {
					showUsage = true;
				}
```
I added "hw |= 1;" and after the check for height: "hw |= 2;"
With this we have only to check if hw contain the value 0 (no -h -w arguments) or 3 (both parameter set).
At the end of the loop we can add:

`if ( hw && hw !=3 ) showUsage = true;`

With this we check if all needed parameters are given and if not so, print the usage and exit.

BenBE:
[...]

PR #148, opened 2022-08-26
new geometry handling / scaling, #146 possible fix

phlash:
As discussed, this introduces new CLI switches for setting capture and virtual camera geometry, re-arranges the order of processing to avoid crashes due to invalid geometries and displays appropriate errors/warnings if deprecated and replacement switches are both used, if geometries change due to the capture device, or aspect ratios do not match.

### DIFF
--- a/app/deepseg.cc
+++ b/app/deepseg.cc
@@ -12,6 +12,7 @@
 #include <istream>
 #include <regex>
 #include <optional>
+#include <utility>
 #include <condition_variable>
 
 #include <opencv2/core/core.hpp>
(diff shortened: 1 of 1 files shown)


app/deepseg.cc
```diff
@@ -461,19 +493,30 @@ int main(int argc, char* argv[]) try {
 		}
 	}
 
+	// prevent use of both deprecated and current switches
+	if (setWorH && (capGeo || vidGeo)) {
+		showUsage = true;
+		fprintf(stderr, "Error: (DEPRECATED) -w/-h used in conjunction with --cg/--vg.\n");
```
BenBE, line 499: Maybe use `perror` instead?

BenBE (approved):
One minor note, apart from that LGTM. Not tested.

app/deepseg.cc
phlash: Only useful if `errno` has been set (which won't happen in this case).

Issue #149, opened 2022-08-27
--cg --vg resize: image streched


jjsarton:
If we don't have the same aspect ratio for the camera and the video output the result is not very pretty.
For this case, we can calculate the region we will scale. For example, if --cg is 640x480 and --vg is 480x480, we my calculate the region to be processed and set
```
cv::Rect crop_region(80, 0,480, 480);
```
and call

```
cv::resize(raw(crop_region), raw, cv::Size(vidGeo.value().first,vidGeo.value().second));
```
instead of

```
cv::resize(raw raw, cv::Size(vidGeo.value().first,vidGeo.value().second));
```
This allows to get a non distorted output.

BenBE:
Can you do a PR to perform the streching/cropping based on a command line switch? TIA.

jjsarton:
Assume that we call backscrub with the arguments:
backscrub -b img.jpg -c /dev/video0 -v /dev/video3 --cg 640x480 --vg 480x480

The raw image (captured by the camera has an aspect ratio 4:3 the output image an aspect ratio of 1:1. Actually, the whole content of the raw image is resized so that the complete image fit to the wanted output size.

This mean that a circle will, after resizing, will be an oval. The height is. for this case bigger as the width.

If we take into account only a centered area and crop first the raw image, a circle will remain a circle.

jjsarton:
I have modified deepseg.c as follow:

added after the last variable declaration within main()
```
	cv::Rect crop_region(0,0,0,0);
```
Before the mainloop added:

```
	if (expWidth != vidGeo.value().first) {
		if (capGeo.value().first >= vidGeo.value().first && 
			capGeo.value().second >= vidGeo.value().second  {
			float ratio = vidGeo.value().first  / (float)vidGeo.value().second;
			int w = vidGeo.value().first * ratio;
			int h = vidGeo.value().second * ratio;
			if (w >= capGeo.value().first) {
				crop_region.x = (int) ((capGeo.value().first - capGeo.value().second) / 2.0);
				crop_region.y = 0;
			} else {
				crop_region.y = (int) ((capGeo.value().second - capGeo.value().first) / 2.0);
				crop_region.x = 0;
			}
			crop_region.width = vidGeo.value().first;
			crop_region.height = vidGeo.value().second;
		} 
	}
````

The code will only be passed if we are scaling down!

Scaling occur with the following:
```
		// scale to virtual camera geometry (if required)
		if (vidGeo != capGeo) {
[...]

PR #153, opened 2022-09-01
added cropping of capture/background

jjsarton:
Added cropping

### DIFF
--- a/app/background.cc
+++ b/app/background.cc
@@ -8,7 +8,7 @@
 #include <opencv2/videoio.hpp>
 #include <opencv2/imgproc.hpp>
 #include <opencv2/highgui.hpp>
-
+#include <lib/libbackscrub.h>
 // Internal state of background processing
 struct background_t {
     int debug;
@@ -183,11 +183,14 @@ int grab_background(std::shared_ptr<background_t> pbkd, int width, int height, c
     if (pbkd->video) {
         // grab frame & frame no. under mutex
         std::unique_lock<std::mutex> hold(pbkd->rawmux);
-        cv::resize(pbkd->raw, out, cv::Size(width, height));
+	cv::Rect_<int> crop = calcCropping(pbkd->raw.cols,pbkd->raw.rows,width, height);
+        cv::resize(pbkd->raw(crop), out, cv::Size(width, height));
         frm = pbkd->frame;
     } else {
         // resize still image as requested into out
-        cv::resize(pbkd->raw, out, cv::Size(width, height));
+	cv::Rect_<int> crop = calcCropping(pbkd->raw.cols,pbkd->raw.rows,width, height);
+	cv::resize(pbkd->raw(crop), out, cv::Size(width, height));
+	out = pbkd->raw;
         frm = 1;
     }
     return frm;
--- a/app/deepseg.cc
+++ b/app/deepseg.cc
@@ -372,6 +372,7 @@ int main(int argc, char* argv[]) try {
 	bool flipVertical = false;
 	int fourcc = 0;
 	size_t blur_strength = 0;
+	cv::Rect_<int> crop_region(0,0,0,0);
 
 	const char* modelname = "selfiesegmentation_mlkit-256x256-2021_01_19-v1215.f16.tflite";
 
@@ -568,6 +569,15 @@ int main(int argc, char* argv[]) try {
 	if (expWidth != vidGeo.value().first) {
 		fprintf(stderr, "Warning: virtual camera aspect ratio does not match capture device.\n");
 	}
+	// calculate crop region, only if result always smaller
+	if (expWidth != vidGeo.value().first &&
+		vidGeo.value().first <= capGeo.value().first &&
+		vidGeo.value().second <= capGeo.value().second) {
+			crop_region = calcCropping(capGeo.value().first,
+			                           capGeo.value().second,
+			                           vidGeo.value().first,
+			                           vidGeo.value().second);
+	}
 
 	// dump settings..
 	printf("debug:  %d\n", debug);
@@ -600,7 +610,12 @@ int main(int argc, char* argv[]) try {
 		}
 	}
 	// default green screen background (at capture true geometry)
-	cv::Mat bg = cv::Mat(capGeo.value().second, capGeo.value().first, CV_8UC3, cv::Scalar(0, 255, 0));
+	cv::Mat bg;
+	if ( crop_region.height == 0) {
+		bg = cv::Mat(capGeo.value().second, capGeo.value().first, CV_8UC3, cv::Scalar(0, 255, 0));
+	} else {
+		bg = cv::Mat(crop_region.width, crop_region.height, CV_8UC3, cv::Scalar(0, 255, 0));
+	}
 
 	// Virtual camera (at specified geometry)
 	int lbfd = loopback_init(s_vcam, vidGeo.value().first, vidGeo.value().second, debug);
@@ -615,9 +630,26 @@ int main(int argc, char* argv[]) try {
 
 
 	// Processing components, all at capture true geometry
-	cv::Mat mask(capGeo.value().second, capGeo.value().first, CV_8U);
+    cv::Mat mask;
+    if ( crop_region.height ) {
+		cv::Mat masktmp((int)(crop_region.height), (int)(crop_region.width), CV_8U);
+		mask = masktmp;
+	} else {
+		cv::Mat masktmp(capGeo.value().second, capGeo.value().first, CV_8U);
+		mask = masktmp;
+	}
+
 	cv::Mat raw;
-	CalcMask ai(s_model.value(), threads, capGeo.value().first, capGeo.value().second);
+	int aiw,aih;
+	if ( crop_region.width == 0) {
+		aiw=capGeo.value().first;
+		aih=capGeo.value().second;
+	} else {
+		aiw=crop_region.width;
+		aih=crop_region.height;
+	}
+	CalcMask ai(s_model.value(), threads, aiw, aih);
+
 	ti.lastns = timestamp();
 	printf("Startup: %ldns\n", diffnanosecs(ti.lastns,ti.bootns));
 
@@ -631,10 +663,14 @@ int main(int argc, char* argv[]) try {
 		// copy new frame to buffer
 		cap.retrieve(raw);
 		ti.retrns = timestamp();
-		ai.set_input_frame(raw);
 		ti.copyns = timestamp();
 
 		if (raw.rows == 0 || raw.cols == 0) continue; // sanity check
+
+		if ( crop_region.height) {
+			raw((cv::Rect_<int>)crop_region).copyTo(raw);
+		}
+		ai.set_input_frame(raw);
 
 		if (filterActive) {
 			// do background detection magic
@@ -646,7 +682,15 @@ int main(int argc, char* argv[]) try {
 			// - default green (initial value)
 			bool canBlur = false;
 			if (pbk) {
-				if (grab_background(pbk, capGeo.value().first, capGeo.value().second, bg)<0)
+				int tw,th;
+				if ( crop_region.height ) {
+					tw = crop_region.width;
+					th = crop_region.height;
+				} else {
+					tw = capGeo.value().first;
+					th = capGeo.value().second;
+				}
+				if (grab_background(pbk, tw, th, bg)<0)
 					throw "Failed to read background frame";
 				canBlur = true;
 			} else if (blur_strength) {
--- a/lib/libbackscrub.cc
+++ b/lib/libbackscrub.cc
@@ -365,7 +365,11 @@ bool bs_maskgen_process(void *context, cv::Mat &frame, cv::Mat &mask) {
 
 	// scale up into full-sized mask
 	cv::Mat tmpbuf;
-	cv::resize(ctx.ofinal(ctx.in_roidim),tmpbuf,ctx.mroi.size());
+	// with body-pix-float-050-8.tflite the size of ctx.ofinal is 33x33
+	// and the wanted roi may be greater as 33x33 so we can crash with
+	// cv::resize(ctx.ofinal(ctx.in_roidim),tmpbuf,ctx.mroi.size());
+	ctx.ofinal.copyTo(tmpbuf);
+	cv::resize(tmpbuf,tmpbuf,ctx.mroi.size());
 
 	// blur at full size for maximum smoothness
 	cv::blur(tmpbuf,ctx.mroi,ctx.blur);
@@ -375,3 +379,22 @@ bool bs_maskgen_process(void *context, cv::Mat &frame, cv::Mat &mask) {
 	return true;
 }
 
+cv::Rect calcCropping(int cw, int ch, int vw, int vh)
+{
+	// if the input and output aspect reatio are not the same
+	// we can crop the source image. For example if the
+	// input image has a 16:9 (1280x720) ratio and the output is 4:3 (960x720)
+	// we will return the cropRegion set as x=160, width=960, y=0, height=720
+	// which the centered part ofe the original image
+	cv::Rect cropRegion = {0,0,0,0};
+	float sc = (float)vw / cw;
+	float st = (float)vh / ch;
+	sc = st > sc ? st : sc;
+	int sz = (int)(vw / sc) - cw;
+	cropRegion.x =  (sz < 0 ? -sz : sz)/2;
+	sz = (int)(vh / sc) - ch;
+	cropRegion.y =  (sz < 0 ? -sz : sz)/2;
+	cropRegion.width = cw - cropRegion.x*2;
+	cropRegion.height = ch - cropRegion.y*2;
+	return cropRegion;
+}
--- a/lib/libbackscrub.h
+++ b/lib/libbackscrub.h
@@ -38,4 +38,6 @@ extern void bs_maskgen_delete(void *context);
 // Process a video frame into a mask
 extern bool bs_maskgen_process(void *context, cv::Mat& frame, cv::Mat &mask);
 
+extern cv::Rect calcCropping(int inWidth, int inHeight, int targetWidth, int targetHight);
+
 #endif


app/background.cc
```diff
@@ -8,7 +8,7 @@
 #include <opencv2/videoio.hpp>
 #include <opencv2/imgproc.hpp>
 #include <opencv2/highgui.hpp>
-
+#include <lib/libbackscrub.h>
```
BenBE, line 11:
```suggestion

#include <lib/libbackscrub.h>

```

app/background.cc
```diff
@@ -183,11 +183,14 @@ int grab_background(std::shared_ptr<background_t> pbkd, int width, int height, c
     if (pbkd->video) {
         // grab frame & frame no. under mutex
         std::unique_lock<std::mutex> hold(pbkd->rawmux);
-        cv::resize(pbkd->raw, out, cv::Size(width, height));
+	cv::Rect_<int> crop = calcCropping(pbkd->raw.cols,pbkd->raw.rows,width, height);
```
BenBE, line 186: Please conserve whitespace style:
```suggestion
        cv::Rect_<int> crop = calcCropping(pbkd->raw.cols,pbkd->raw.rows,width, height);
```

app/background.cc
```diff
@@ -183,11 +183,14 @@ int grab_background(std::shared_ptr<background_t> pbkd, int width, int height, c
...
+        cv::resize(pbkd->raw(crop), out, cv::Size(width, height));
         frm = pbkd->frame;
     } else {
         // resize still image as requested into out
-        cv::resize(pbkd->raw, out, cv::Size(width, height));
+	cv::Rect_<int> crop = calcCropping(pbkd->raw.cols,pbkd->raw.rows,width, height);
+	cv::resize(pbkd->raw(crop), out, cv::Size(width, height));
+	out = pbkd->raw;
```
BenBE, line 193:
```suggestion
        cv::Rect_<int> crop = calcCropping(pbkd->raw.cols, pbkd->raw.rows, width, height);
        cv::resize(pbkd->raw(crop), out, cv::Size(width, height));
        out = pbkd->raw;
```

Sure it's `cv::Rect_` instead of `cv::Rect`? Can't test right now.

app/deepseg.cc
```diff
@@ -372,6 +372,7 @@ int main(int argc, char* argv[]) try {
 	bool flipVertical = false;
 	int fourcc = 0;
 	size_t blur_strength = 0;
+	cv::Rect_<int> crop_region(0,0,0,0);
```
BenBE, line 375:
```suggestion
	cv::Rect_<int> crop_region(0, 0, 0, 0);
```

app/deepseg.cc
```diff
@@ -568,6 +569,15 @@ int main(int argc, char* argv[]) try {
...
 		fprintf(stderr, "Warning: virtual camera aspect ratio does not match capture device.\n");
 	}
+	// calculate crop region, only if result always smaller
+	if (expWidth != vidGeo.value().first &&
+		vidGeo.value().first <= capGeo.value().first &&
+		vidGeo.value().second <= capGeo.value().second) {
+			crop_region = calcCropping(capGeo.value().first,
+			                           capGeo.value().second,
+			                           vidGeo.value().first,
+			                           vidGeo.value().second);
```
BenBE, line 579: Avoid too long overhangs. Indentation should only increase one level, not those alignments to under the opening parens (which just look ugly).

```suggestion
			crop_region = calcCropping(
				capGeo.value().first, capGeo.value().second,
				vidGeo.value().first, vidGeo.value().second);
```

Having this repeated `var.value().field` kinda looks ugly, even though it is right semantically …

app/deepseg.cc
```diff
@@ -600,7 +610,12 @@ int main(int argc, char* argv[]) try {
 		}
 	}
 	// default green screen background (at capture true geometry)
-	cv::Mat bg = cv::Mat(capGeo.value().second, capGeo.value().first, CV_8UC3, cv::Scalar(0, 255, 0));
+	cv::Mat bg;
+	if ( crop_region.height == 0) {
+		bg = cv::Mat(capGeo.value().second, capGeo.value().first, CV_8UC3, cv::Scalar(0, 255, 0));
+	} else {
+		bg = cv::Mat(crop_region.width, crop_region.height, CV_8UC3, cv::Scalar(0, 255, 0));
+	}
```
BenBE, line 618: DRY:
```suggestion
	std::pair<int, int> bg_dim = !crop_region.height ?
		{ capGeo.value().second, capGeo.value().first } :
		{ crop_region.width, crop_region.height };
	cv::Mat bg = cv::Mat(bg_dim.second, bg_dim.first, CV_8UC3, cv::Scalar(0, 255, 0));
```

app/deepseg.cc
```diff
@@ -615,9 +630,26 @@ int main(int argc, char* argv[]) try {
...
 	// Processing components, all at capture true geometry
-	cv::Mat mask(capGeo.value().second, capGeo.value().first, CV_8U);
+    cv::Mat mask;
+    if ( crop_region.height ) {
+		cv::Mat masktmp((int)(crop_region.height), (int)(crop_region.width), CV_8U);
+		mask = masktmp;
+	} else {
+		cv::Mat masktmp(capGeo.value().second, capGeo.value().first, CV_8U);
+		mask = masktmp;
+	}
```
BenBE, line 640: Optimize similar to example shown above. `masktmp` is unused (except for assignment). Drop unnecessary parens.

app/deepseg.cc
```diff
@@ -615,9 +630,26 @@ int main(int argc, char* argv[]) try {
...
 	cv::Mat raw;
-	CalcMask ai(s_model.value(), threads, capGeo.value().first, capGeo.value().second);
+	int aiw,aih;
+	if ( crop_region.width == 0) {
+		aiw=capGeo.value().first;
+		aih=capGeo.value().second;
+	} else {
+		aiw=crop_region.width;
+		aih=crop_region.height;
+	}
```
BenBE, line 650:
```suggestion
	int aiw, aih;
	std::tie(aiw, aih) = !crop_region.width ?
		{ capGeo.value().first, capGeo.value().second } :
		{ crop_region.width, crop_region.height };
```

app/deepseg.cc
```diff
@@ -631,11 +663,15 @@ int main(int argc, char* argv[]) try {
...
 		ti.retrns = timestamp();
-		ai.set_input_frame(raw);
 		ti.copyns = timestamp();
 
 		if (raw.rows == 0 || raw.cols == 0) continue; // sanity check
 
+		if ( crop_region.height) {
+			raw((cv::Rect_<int>)crop_region).copyTo(raw);
+		}
+		ai.set_input_frame(raw);
```
BenBE, line 673:
```suggestion

		ai.set_input_frame(raw);
```

Just a bit more space for visual code structure …

app/deepseg.cc
```diff
@@ -646,7 +682,15 @@ int main(int argc, char* argv[]) try {
...
 			if (pbk) {
-				if (grab_background(pbk, capGeo.value().first, capGeo.value().second, bg)<0)
+				int tw,th;
+				if ( crop_region.height ) {
+					tw = crop_region.width;
+					th = crop_region.height;
+				} else {
+					tw = capGeo.value().first;
+					th = capGeo.value().second;
+				}
```
BenBE, line 692: Should use `std::tie` as shown above.

app/deepseg.cc
```diff
@@ -646,7 +682,15 @@ int main(int argc, char* argv[]) try {
...
+				if (grab_background(pbk, tw, th, bg)<0)
```
BenBE, line 693:
```suggestion
				if (grab_background(pbk, tw, th, bg) < 0)
```

lib/libbackscrub.cc
```diff
@@ -365,7 +365,11 @@ bool bs_maskgen_process(void *context, cv::Mat &frame, cv::Mat &mask) {
 
 	// scale up into full-sized mask
 	cv::Mat tmpbuf;
-	cv::resize(ctx.ofinal(ctx.in_roidim),tmpbuf,ctx.mroi.size());
+	// with body-pix-float-050-8.tflite the size of ctx.ofinal is 33x33
+	// and the wanted roi may be greater as 33x33 so we can crash with
+	// cv::resize(ctx.ofinal(ctx.in_roidim),tmpbuf,ctx.mroi.size());
+	ctx.ofinal.copyTo(tmpbuf);
+	cv::resize(tmpbuf,tmpbuf,ctx.mroi.size());
```
BenBE, line 372:
```suggestion
	cv::resize(tmpbuf, tmpbuf, ctx.mroi.size());
```

lib/libbackscrub.cc
```diff
@@ -375,3 +379,22 @@ bool bs_maskgen_process(void *context, cv::Mat &frame, cv::Mat &mask) {
 	return true;
 }
 
+cv::Rect calcCropping(int cw, int ch, int vw, int vh)
+{
+	// if the input and output aspect reatio are not the same
+	// we can crop the source image. For example if the
+	// input image has a 16:9 (1280x720) ratio and the output is 4:3 (960x720)
+	// we will return the cropRegion set as x=160, width=960, y=0, height=720
+	// which the centered part ofe the original image
```
BenBE, line 388:
```suggestion
	// which is the centered part of the original image
```

lib/libbackscrub.cc
```diff
@@ -375,3 +379,22 @@ bool bs_maskgen_process(void *context, cv::Mat &frame, cv::Mat &mask) {
 	return true;
 }
 
+cv::Rect calcCropping(int cw, int ch, int vw, int vh)
+{
+	// if the input and output aspect reatio are not the same
```
BenBE, line 384:
```suggestion
	// if the input and output aspect ratio are not the same
```

lib/libbackscrub.cc
```diff
@@ -375,3 +379,22 @@ bool bs_maskgen_process(void *context, cv::Mat &frame, cv::Mat &mask) {
...
+	// we can crop the source image. For example if the
+	// input image has a 16:9 (1280x720) ratio and the output is 4:3 (960x720)
+	// we will return the cropRegion set as x=160, width=960, y=0, height=720
+	// which the centered part ofe the original image
+	cv::Rect cropRegion = {0,0,0,0};
```
BenBE, line 389:
```suggestion
	cv::Rect cropRegion = {0, 0, 0, 0};
```

lib/libbackscrub.cc
```diff
@@ -375,3 +379,22 @@ bool bs_maskgen_process(void *context, cv::Mat &frame, cv::Mat &mask) {
...
+	float sc = (float)vw / cw;
+	float st = (float)vh / ch;
+	sc = st > sc ? st : sc;
+	int sz = (int)(vw / sc) - cw;
+	cropRegion.x =  (sz < 0 ? -sz : sz)/2;
+	sz = (int)(vh / sc) - ch;
+	cropRegion.y =  (sz < 0 ? -sz : sz)/2;
+	cropRegion.width = cw - cropRegion.x*2;
+	cropRegion.height = ch - cropRegion.y*2;
```
BenBE, line 398:
```suggestion

	int sx = (int)(vw / sc) - cw;
	cropRegion.x =  (sx < 0 ? -sx : sx) / 2;
	int sy = (int)(vh / sc) - ch;
	cropRegion.y =  (sy < 0 ? -sy : sy) / 2;

	cropRegion.width = cw - cropRegion.x * 2;
	cropRegion.height = ch - cropRegion.y * 2;
```

BenBE (changes requested):
Couldn't test things yet, but those are what I noticed while skimming over.

app/deepseg.cc
```diff
@@ -631,11 +663,15 @@ int main(int argc, char* argv[]) try {
 		// copy new frame to buffer
 		cap.retrieve(raw);
 		ti.retrns = timestamp();
-		ai.set_input_frame(raw);
 		ti.copyns = timestamp();
```
BenBE, line 666: This line should stay with the previous one. Just move it down to after line 673 too …

app/deepseg.cc
jjsarton: There is a mistake here! I must check this!

app/deepseg.cc
jjsarton: See previous not committed suggestion.

app/deepseg.cc
BenBE: I hoped the compiler can resolve the initializer list, but you might need to make it explicit as `std::make_pair(a, b)`.

app/deepseg.cc
jjsarton: I think we will have the same problems as for the previous not committed suggestions.

app/deepseg.cc
jjsarton: I have tried to solve this, but there is no advantage compared to my first code, only more code.

app/background.cc
```diff
@@ -183,11 +185,14 @@ int grab_background(std::shared_ptr<background_t> pbkd, int width, int height, c
     if (pbkd->video) {
         // grab frame & frame no. under mutex
         std::unique_lock<std::mutex> hold(pbkd->rawmux);
-        cv::resize(pbkd->raw, out, cv::Size(width, height));
+        cv::Rect_<int> crop = calcCropping(pbkd->raw.cols, pbkd->raw.rows, width, height);
```
BenBE, line 188:
```suggestion
        cv::Rect crop = calcCropping(pbkd->raw.cols, pbkd->raw.rows, width, height);
```

app/background.cc
```diff
@@ -183,11 +185,14 @@ int grab_background(std::shared_ptr<background_t> pbkd, int width, int height, c
...
+        cv::resize(pbkd->raw(crop), out, cv::Size(width, height));
         frm = pbkd->frame;
     } else {
         // resize still image as requested into out
-        cv::resize(pbkd->raw, out, cv::Size(width, height));
+        cv::Rect_<int> crop = calcCropping(pbkd->raw.cols, pbkd->raw.rows, width, height);
```
BenBE, line 193:
```suggestion
        cv::Rect crop = calcCropping(pbkd->raw.cols, pbkd->raw.rows, width, height);
```

app/deepseg.cc
```diff
@@ -613,11 +627,27 @@ int main(int argc, char* argv[]) try {
...
 	// Processing components, all at capture true geometry
-	cv::Mat mask(capGeo.value().second, capGeo.value().first, CV_8U);
+	cv::Mat mask;
+	if ( crop_region.height ) {
+		cv::Mat masktmp((int)(crop_region.height), (int)(crop_region.width), CV_8U);
+		mask = masktmp;
+	} else {
+		cv::Mat masktmp(capGeo.value().second, capGeo.value().first, CV_8U);
+		mask = masktmp;
+	}
```
BenBE, line 638:
```suggestion
	if (crop_region.height) {
		mask = cv::Mat((int)crop_region.height, (int)crop_region.width, CV_8U);
	} else {
		mask = cv::Mat(capGeo->second, capGeo->first, CV_8U);
	}
```

app/deepseg.cc
```diff
@@ -631,22 +661,35 @@ int main(int argc, char* argv[]) try {
 		// copy new frame to buffer
 		cap.retrieve(raw);
 		ti.retrns = timestamp();
+
+		if (raw.rows == 0 || raw.cols == 0) continue; // sanity check
+
+		if ( crop_region.height) {
+			raw((cv::Rect_<int>)crop_region).copyTo(raw);
+		}
```
BenBE, line 669:
```suggestion
		if (crop_region.height) {
			raw(crop_region).copyTo(raw);
		}
```

BenBE (changes requested):
`capGeo.value().first`/`capGeo.value().second` can be shortened to `capGeo->first`/`capGeo->second`

app/deepseg.cc
jjsarton: This may possibly be:
```
	std::pair<int, int> bg_dim = {(int)capGeo.value().first, (int)capGeo.value().second};
	if ( crop_region.height ) {
		bg_dim = {crop_region.width, crop_region.height};
	}
	cv::Mat mask(bg_dim.value().first, bg_dim.value().second, CV_8U);
```

BenBE: Replace `std::pair<size_t, size_t>` by `cv::Size` and the overall code will magically shrink quite a bit …

jjsarton:
On my fork jjsarton/backscrub the origin branch contain the latest code including your suggestions or the revised code.

I thing that your suggestion concerning line 680 to 687 (grab_background) may not be changed we pass 2 variables and don't need to use a class in order to put them the variables into a class and extracting them later!

app/background.cc
```diff
@@ -183,11 +185,14 @@ int grab_background(std::shared_ptr<background_t> pbkd, int width, int height, c
...
         // grab frame & frame no. under mutex
         std::unique_lock<std::mutex> hold(pbkd->rawmux);
-        cv::resize(pbkd->raw, out, cv::Size(width, height));
+        cv::Rect_<int> crop = calcCropping(pbkd->raw.cols,pbkd->raw.rows,width,height);
+        cv::resize(pbkd->raw(crop), out, cv::Size(width, height));
         frm = pbkd->frame;
     } else {
         // resize still image as requested into out
-        cv::resize(pbkd->raw, out, cv::Size(width, height));
+        cv::Rect_<int> crop = calcCropping(pbkd->raw.cols,pbkd->raw.rows,width,height);
```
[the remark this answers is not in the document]
jjsarton: I have updates my cropping branch accordingly. Your suggestion is within my code

app/deepseg.cc
```diff
@@ -372,6 +372,7 @@ int main(int argc, char* argv[]) try {
 	bool flipVertical = false;
 	int fourcc = 0;
 	size_t blur_strength = 0;
+	cv::Rect_<int> crop_region(0, 0, 0, 0);
```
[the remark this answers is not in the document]
jjsarton: I have updates my cropping branch accordingly. Your suggestion is within my code

app/background.cc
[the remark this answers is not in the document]
jjsarton: I have updates my cropping branch accordingly. Your suggestion is within my code

app/background.cc
jjsarton: I have updates my cropping branch accordingly. Your suggestion is within my code

app/background.cc
jjsarton: I have updates my cropping branch accordingly. Your suggestion is within my code

app/deepseg.cc
jjsarton: I have updates my cropping branch accordingly. Your suggestion is within my code

app/deepseg.cc
jjsarton: This is modified line 633 as:

```
	std::optional<std::pair<size_t, size_t>> mask_dim = capGeo;
	if ( crop_region.height ) {
		mask_dim = {crop_region.width, crop_region.height};
	}
	cv::Mat mask(mask_dim.value().second, mask_dim.value().first, CV_8U);``
```
within my cropping branch

app/deepseg.cc
jjsarton: corrected within my branch

app/deepseg.cc
jjsarton: Actually the code is:

```
		// copy new frame to buffer
		cap.retrieve(raw);
		ti.retrns = timestamp();

		if (raw.rows == 0 || raw.cols == 0) continue; // sanity check

		if ( crop_region.height) {
			raw(crop_region).copyTo(raw);
		}
		ai.set_input_frame(raw);
		ti.copyns = timestamp();

		// do background detection magic
```
I hope that moving the times are set correctl.

app/background.cc
```diff
@@ -183,11 +185,14 @@ int grab_background(std::shared_ptr<background_t> pbkd, int width, int height, c
     if (pbkd->video) {
         // grab frame & frame no. under mutex
         std::unique_lock<std::mutex> hold(pbkd->rawmux);
-        cv::resize(pbkd->raw, out, cv::Size(width, height));
+        cv::Rect_<int> crop = calcCropping(pbkd->raw.cols,pbkd->raw.rows,width,height);
```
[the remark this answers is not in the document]
jjsarton: modified

app/deepseg.cc
jjsarton: This is modified like the code for line 633

app/deepseg.cc
BenBE: Note the `.value().first` -> `->first` change in my suggestion. Saves quite a bit of typing.

app/deepseg.cc
```diff
@@ -568,6 +569,14 @@ int main(int argc, char* argv[]) try {
...
 		fprintf(stderr, "Warning: virtual camera aspect ratio does not match capture device.\n");
 	}
+	// calculate crop region, only if result always smaller
+	if (expWidth != vidGeo.value().first &&
+		vidGeo.value().first <= capGeo.value().first &&
+		vidGeo.value().second <= capGeo.value().second) {
+			crop_region = calcCropping(
+				capGeo.value().first, capGeo.value().second,
+				vidGeo.value().first, vidGeo.value().second);
+	}
```
BenBE, line 579:
```suggestion
	if (expWidth != vidGeo->first &&
		vidGeo->first <= capGeo->first &&
		vidGeo->second <= capGeo->second
	) {
		crop_region = calcCropping(
			capGeo->first, capGeo->second,
			vidGeo->first, vidGeo->second);
	}
```

app/deepseg.cc
```diff
@@ -600,7 +609,11 @@ int main(int argc, char* argv[]) try {
 		}
 	}
 	// default green screen background (at capture true geometry)
-	cv::Mat bg = cv::Mat(capGeo.value().second, capGeo.value().first, CV_8UC3, cv::Scalar(0, 255, 0));
+	std::optional<std::pair<size_t, size_t>> bg_dim = capGeo;
+	if ( crop_region.height == 0) {
+		bg_dim = {crop_region.width, crop_region.height};
+	}
+	cv::Mat bg(bg_dim.value().second, bg_dim.value().first, CV_8UC3, cv::Scalar(0, 255, 0));
```
BenBE, line 616:
```suggestion
	std::pair<size_t, size_t> bg_dim = *capGeo;
	if (!crop_region.height) {
		bg_dim = {crop_region.width, crop_region.height};
	}
	cv::Mat bg(bg_dim.second, bg_dim.first, CV_8UC3, cv::Scalar(0, 255, 0));
```

But actually using `cv::Size` for bg_dim would be semantically much more sane, than abusing `std::pair`.

app/deepseg.cc
```diff
@@ -613,11 +626,24 @@ int main(int argc, char* argv[]) try {
...
 	});
 
-
 	// Processing components, all at capture true geometry
-	cv::Mat mask(capGeo.value().second, capGeo.value().first, CV_8U);
+	std::optional<std::pair<size_t, size_t>> mask_dim = capGeo;
+	if ( crop_region.height ) {
+		mask_dim = {crop_region.width, crop_region.height};
+	}
+	cv::Mat mask(mask_dim.value().second, mask_dim.value().first, CV_8U);
```
BenBE, line 634:
```suggestion
	std::pair<size_t, size_t> mask_dim = *capGeo;
	if (crop_region.height) {
		mask_dim = {crop_region.width, crop_region.height};
	}
	cv::Mat mask(mask_dim.second, mask_dim.first, CV_8U);
```

Same remark as above re `cv::Size`.

app/deepseg.cc
```diff
@@ -613,11 +626,24 @@ int main(int argc, char* argv[]) try {
...
+
 	cv::Mat raw;
-	CalcMask ai(s_model.value(), threads, capGeo.value().first, capGeo.value().second);
+	int aiw,aih;
+	if ( crop_region.width == 0) {
+		aiw=capGeo.value().first;
+		aih=capGeo.value().second;
```
BenBE, line 640:
```suggestion
		aiw=capGeo->first;
		aih=capGeo->second;
```

app/deepseg.cc
```diff
@@ -613,11 +626,24 @@ int main(int argc, char* argv[]) try {
...
+	std::optional<std::pair<size_t, size_t>> mask_dim = capGeo;
+	if ( crop_region.height ) {
+		mask_dim = {crop_region.width, crop_region.height};
+	}
+	cv::Mat mask(mask_dim.value().second, mask_dim.value().first, CV_8U);
+
 	cv::Mat raw;
-	CalcMask ai(s_model.value(), threads, capGeo.value().first, capGeo.value().second);
+	int aiw,aih;
+	if ( crop_region.width == 0) {
```
BenBE, line 638:
```suggestion
	if (!crop_region.width) {
```

app/deepseg.cc
```diff
@@ -613,11 +626,24 @@ int main(int argc, char* argv[]) try {
...
+		aiw=capGeo.value().first;
+		aih=capGeo.value().second;
+	} else {
+		aiw=crop_region.width;
+		aih=crop_region.height;
+	}
+	CalcMask ai(s_model.value(), threads, aiw, aih);
```
BenBE, line 645:
```suggestion
	CalcMask ai(*s_model, threads, aiw, aih);
```

app/deepseg.cc
```diff
@@ -631,22 +657,35 @@ int main(int argc, char* argv[]) try {
 		// copy new frame to buffer
 		cap.retrieve(raw);
 		ti.retrns = timestamp();
+
+		if (raw.rows == 0 || raw.cols == 0) continue; // sanity check
+
+		if ( crop_region.height) {
```
BenBE, line 663:
```suggestion
		if (crop_region.height) {
```

app/deepseg.cc
```diff
@@ -631,22 +657,35 @@ int main(int argc, char* argv[]) try {
...
 			bool canBlur = false;
 			if (pbk) {
-				if (grab_background(pbk, capGeo.value().first, capGeo.value().second, bg)<0)
+				int tw,th;
+				if ( crop_region.height ) {
+					tw = crop_region.width;
+					th = crop_region.height;
+				} else {
+					tw = capGeo.value().first;
+					th = capGeo.value().second;
```
BenBE, line 686:
```suggestion
					tw = capGeo->first;
					th = capGeo->second;
```

BenBE (changes requested):

app/deepseg.cc
jjsarton: I had a look at the already provided code and used the same syntax. I had not looked correctly to your suggestion. I will try this.

jjsarton:
Github is very slow!
I had committed the suggested changes and compared the own source with my branch. All modifications what not take into account!
I have pushed my own file to github, looked again on the push request and seen that some suggestion (which what no more shown while I was within the commit process) were already present. Sorry!

BenBE:
Looks already good so far. Care to squash/fixup those commits into one (or few) …

jjsarton:
I have to learn how to deal with GitHub! I hope that I will learn this.

jjsarton:
@BenBE 
I have done more checks and had to correct a little bug within app/background.cc (line 195). The fix is committed within my cropping branch.

app/background.cc
```diff
@@ -191,7 +191,10 @@ int grab_background(std::shared_ptr<background_t> pbkd, int width, int height, c
     } else {
         // resize still image as requested into out
         cv::Rect crop = calcCropping(pbkd->raw.cols, pbkd->raw.rows, width, height);
-        cv::resize(pbkd->raw(crop), pbkd->raw, cv::Size(width, height));
+		// under some circumstances we must do the job with 2 step!
+		// without this resize(pbkd->raw(crop), pbkd->raw, ...) may fail.
```
BenBE, line 195:
```suggestion
	// Under some circumstances we must do the job in two steps!
	// Otherwise this resize(pbkd->raw(crop), pbkd->raw, ...) may fail.
```

jjsarton: I have pushed the suggestion and also try to reproduce the same problem for the video as with the still image for the video part. I can produce a crash only on very strange resolutions for the video and the output size, and the code within background what never involved. So I expect that a normal user will not have issue with the actual code.

jjsarton:
I have a little error at line 613 deepseg.c and fixed it. I have also removes 2 space within an if query as you recommend it.
I have also pushed this to my origin branch.

lib/libbackscrub.cc
```diff
@@ -375,3 +379,25 @@ bool bs_maskgen_process(void *context, cv::Mat &frame, cv::Mat &mask) {
 	return true;
 }
 
+cv::Rect calcCropping(int cw, int ch, int vw, int vh) {
```
phlash, line 382: May I suggest:
 * use the same variable names as per the prototype in the header, they are more meaningful
 * stick to our library naming convention so `bs_calc_cropping` or similar.

lib/libbackscrub.h
```diff
@@ -38,4 +38,6 @@ extern void bs_maskgen_delete(void *context);
 // Process a video frame into a mask
 extern bool bs_maskgen_process(void *context, cv::Mat& frame, cv::Mat &mask);
 
+extern cv::Rect calcCropping(int inWidth, int inHeight, int targetWidth, int targetHight);
```
phlash, line 41: As per the code comment - please stick to existing naming convention of `bs_xxx` :)

lib/libbackscrub.cc
```diff
@@ -365,7 +365,11 @@ bool bs_maskgen_process(void *context, cv::Mat &frame, cv::Mat &mask) {
 
 	// scale up into full-sized mask
 	cv::Mat tmpbuf;
-	cv::resize(ctx.ofinal(ctx.in_roidim),tmpbuf,ctx.mroi.size());
+	// with body-pix-float-050-8.tflite the size of ctx.ofinal is 33x33
+	// and the wanted roi may be greater as 33x33 so we can crash with
+	// cv::resize(ctx.ofinal(ctx.in_roidim),tmpbuf,ctx.mroi.size());
+	ctx.ofinal.copyTo(tmpbuf);
+	cv::resize(tmpbuf, tmpbuf, ctx.mroi.size());
```
phlash, line 372: Not sure this is correct? Selecting an ROI from the final output is done because we may have centred the frame into the model earlier (line 289), this change removes the selection step. The calculations at line 237 onward should ensure that `in_roidim` _cannot_ be larger than `ofinal` (unless the model itself has output dim < input dim, which I have never seen but I guess could occur?). Perhaps there is a rounding error there that needs fixing instead of this?

app/background.cc
```diff
@@ -183,11 +185,17 @@ int grab_background(std::shared_ptr<background_t> pbkd, int width, int height, c
     if (pbkd->video) {
         // grab frame & frame no. under mutex
         std::unique_lock<std::mutex> hold(pbkd->rawmux);
-        cv::resize(pbkd->raw, out, cv::Size(width, height));
+        cv::Rect crop = calcCropping(pbkd->raw.cols, pbkd->raw.rows, width, height);
```
phlash, line 188: We could lift this calculation out of the processing loop to a structure variable, as sizes will not change during operation.

app/background.cc
```diff
@@ -183,11 +185,17 @@ int grab_background(std::shared_ptr<background_t> pbkd, int width, int height, c
...
+        cv::resize(pbkd->raw(crop), out, cv::Size(width, height));
         frm = pbkd->frame;
     } else {
         // resize still image as requested into out
-        cv::resize(pbkd->raw, out, cv::Size(width, height));
+        cv::Rect crop = calcCropping(pbkd->raw.cols, pbkd->raw.rows, width, height);
```
phlash, line 193: ..as above, could be removed in favour of a state variable, calculated just once.

lib/libbackscrub.cc
jjsarton: I am lazy, but you are right.

lib/libbackscrub.h
jjsarton: Okay

lib/libbackscrub.cc
jjsarton: I modified the original code because I had a crash within opencv.
I ha done this because the model used had an output of 33x33. 
I have already performed a lot of tests and never shown an error here. I will check this again.TT

jjsarton: I have just tested this again. With the old code, using backscrub/models/body-pix-float-050-8.tflite I get again a crash, with my correction this work well.
the real size for ofinal is 33x33 and with the ctx.in_roidim we have a size of 256x256!!!

app/background.cc
jjsarton: I must think about that.

lib/libbackscrub.cc
jjsarton: Done, not pushed

lib/libbackscrub.h
jjsarton: done not pushed

app/background.cc
jjsarton: I must review this part.

jjsarton: I have put the variable bg_stored into the structure background_t.
The code is now:
 ```
    } else {
		if (!pbkd->bg_stored) {
			// resize still image as requested into out
			cv::Rect crop = bs_calc_cropping(pbkd->raw.cols, pbkd->raw.rows, width, height);
			// Under some circumstances we must do the job in two steps!
			// Otherwise this resize(pbkd->raw(crop), pbkd->raw, ...) may fail.
			pbkd->raw(crop).copyTo(pbkd->raw);
			cv::resize(pbkd->raw, pbkd->raw, cv::Size(width, height));
			pbkd->bg_stored = true;
		}
		pbkd->raw = out;
		frm = 1;
	}
    return frm;
```

**Comment by jjsarton** [comment]
I have pushed the modified code to my cropping branch.

File: app/background.cc (hunk 37 above)
Comment (jjsarton) [reply app/background.cc]: see below

File: lib/libbackscrub.cc (hunk 36 above)
Comment (phlash) [reply lib/libbackscrub.cc:372]: Yep - it's because they have an _output stride_ variable in the model which defaults to 32, so we get input dim/32 == 257/32 ~= 33.

I'm happy to leave the crash for now, you are officially "not making it worse" with this PR :grin:, and we'll probably want to revisit the logic in multiple places to fix this properly.

File: app/background.cc (hunk 38 above)
Comment (phlash) [reply app/background.cc]: Don't worry too much about this, it's a minor performance enhancement to avoid a couple of floating point ops per frame, when we are running a whole inference model too.. I'd rather the code was readable than maximum speed :smile:

File: app/background.cc [hunk 39]
```diff
@@ -18,6 +20,7 @@ struct background_t {
     int frame;
     double fps;
     cv::Mat raw;
+    int bg_stored;
     std::mutex rawmux;
```
phlash, line 24: `bool` as it's a true/false indicator?

app/background.cc
```diff
@@ -18,6 +20,7 @@ struct background_t {
     int frame;
     double fps;
     cv::Mat raw;
+    int bg_stored;
```
phlash, line 23: `bool` perhaps as it's used only for true/false?

app/background.cc
```diff
@@ -143,7 +147,7 @@ std::shared_ptr<background_t> load_background(const std::string& path, int debug
         //  if: can read 2 video frames => it's a video
         //  else: is loaded as an image => it's an image
         //  else: it's not usable.
-        if (pbkd->cap.read(pbkd->raw) && pbkd->cap.read(pbkd->raw)) {
+        if (cnt > -1) {
```
phlash, line 150: If I recall, I chose this current method as `cnt` could be > 1 for some image files (multiple resolutions?) but they would not play as a video... please test with all the variations in `backgrounds` folder :smile:

app/background.cc
```diff
@@ -183,13 +187,22 @@ int grab_background(std::shared_ptr<background_t> pbkd, int width, int height, c
...
+			cv::Rect crop = bs_calc_cropping(pbkd->raw.cols, pbkd->raw.rows, width, height);
+			// Under some circumstances we must do the job in two steps!
+			// Otherwise this resize(pbkd->raw(crop), pbkd->raw, ...) may fail.
+			pbkd->raw(crop).copyTo(pbkd->raw);
+			cv::resize(pbkd->raw, pbkd->raw, cv::Size(width, height));
+			pbkd->bg_stored = true;
+		}
+		out = pbkd->raw ;
+		frm = 1;
+	}
```
phlash, line 205: Thank you for fixing this - it was me being lazy to start with... :+1:

phlash:
Thanks for your patience @jjsarton, this is looking good. You might have missed a comment I put on the issue: https://github.com/floe/backscrub/issues/149#issuecomment-1242717372, which refers back to @BenBE asking for the cropping behaviour to be controlled by a switch, as it's opinionated behaviour that others may not want.

app/background.cc
jjsarton: You are right, I have modified this. Initializing was done with false, setting with true.

app/background.cc
jjsarton: I will perform the tests. Using fps which shall be greater than 0.00000 for a video may be better.

app/background.cc
jjsarton: Some files contain spaces, other tabs. background.cc use spaces. I have first replaced my tabs with spaces as indentation within this file.

app/background.cc
jjsarton: Unfortunately, my system don't work with animated.gif. Due to this I can't check if the fps which would be processed on systems allowing this will return a frame rate of 45 Fps (value obtained while converting the file via ffmpeg to a stream.

Jpg file may contain a thumbnail image, do reading twice work, and the file is recognized as video. Reading 3 times will give the expected result, but this shall not be the right way.

The test for picture/video is now based on the fps.  I think that this shall always work.

jjsarton:
Pushed the actual revised version.

BenBE:
Could you squash all the fixup and correction commits into their respective commit of where they should have gone in the first place? You can/should keep commits that introduce/change behavior though. If I'm not too mistaken you should end up at around 5-ish commits. Makes the whole commit history much more reasonable.

FWIW: Each (intermediate) commit should compile on its own.

app/background.cc
phlash: I'll grab your PR and check what I find here in the next day or so - thanks.

jjsarton:
> Could you squash all the fixup and correction commits into their respective commit of where they should have gone in the first place? You can/should keep commits that introduce/change behavior though. If I'm not too mistaken you should end up at around 5-ish commits. Makes the whole commit history much more reasonable.
> 
> FWIW: Each (intermediate) commit should compile on its own.

I will not do this (not meaning full). My cropping directory respect the suggestion done unless there was not okay.
The only point we have to discuss (including the comments) is the recognizing of video or still image. 
I can't get animated.gif working (GIF is a deprecated format) and can't test if we should get the expected 45 fps on a system which allows to process gif file with backscrub.

app/background.cc
phlash: OK, I've checked the behaviour as the current PR has it (`fps > 0`), and it detects _all_ the backgrounds as videos, but does not fail, since the video loop logic resets the position on each request... we may be able to simplify all this to assume video at all times?

BenBE:
> > Could you squash all the fixup and correction commits into their respective commit of where they should have gone in the first place? You can/should keep commits that introduce/change behavior though. If I'm not too mistaken you should end up at around 5-ish commits. Makes the whole commit history much more reasonable.
> > FWIW: Each (intermediate) commit should compile on its own.
> 
> I will not do this (not meaning full). My cropping directory respect the suggestion done unless there was not okay. The only point we have to discuss (including the comments) is the recognizing of video or still image. I can't get animated.gif working (GIF is a deprecated format) and can't test if we should get the expected 45 fps on a system which allows to process gif file with backscrub.

Unfortunately I'm not sure what you were trying to say and how that related to the comment of mine that you quoted.

app/background.cc
jjsarton: I am surprised!
The backsrcrub binary for the cropping branch work as expected, a video is recognized as video and pictures as pictures.

For total_landscaping.jpg I get as output:

```
background properties:
	vid: no
	fcc: 00000000 ()
	fps: 0.000000
	cnt: -1
```
What tell your system? 

I use Fedora 36 XFCE with opencv 4.5.5. What do you use?
May be that we check for fps > 0.001 in order to have the right condition. 

I have tried the following on my development environment:

```
int w1,w2=0;
if(pbkd->cap.read(pbkd->raw)) {
	w1=pbkd->raw.cols;
	if (pbkd->cap.read(pbkd->raw)) {
		w2=pbkd->raw.cols;
	}
}
if(w2 == w1) {
```
This is basically the same as the old condition, but I assume that the width of 2 consecutive images has to be the same.
The final test will pass if we have a video, fail if we have an image, even width thumbnails   (must have a lower size).

The detection was okay on my system.

jjsarton: According to opencv cap.get(cv::CAP_PROP_FRAME_COUNT) return the  'Number of frames in the video file'
The first intention to use the count as flag for video or still image, what okay. Asking for CAP_PROP_FPS shall also work.

Work with images as if they were a video is not a great idea. If the image has a big size, reading, ... will take more time as
Retrieving the thumb as now realized within background.cc (Your suggestion).

phlash: @jjsarton I'm running on Debian stable (11), OpenCV 4.5.1, here's what I get for each background type:
 * `animated.gif` => vid: yes, fps: 45.0.., cnt: 36
 * `background_bauhaus.png` => vid: yes, fps: 25.0.., cnt: -2147483648
 * `rotating_earth.webm` => vid: yes, fps: 30.0.., cnt: 916
 * `total_landscaping.jpg` => vid: yes, fps: 25.0.., cnt: 1

..so this looks like we are heavily dependant on unstable OpenCV behaviour :disappointed:

jjsarton:
I have configured and installed my own opencv version. With this version ffmpeg work and unfortunately, still picture are recognized as video. In order to solve this I modified (on my test version) background.cc, load_background).
After initializing pbkd I have inserted:

```
 size_t imgages = imcount(path, cv::IMREAD_ANYCOLOR);
```
The test video or still picture is now
```
if (!images) {
```
I have tested this against the original installed opencv with don*t support ffmpeg and my own compiled opencv. For both the video detection worked.

I propose that I implement this within my cropping branch. We can also delete the comments concerning the detection or add some if you tell me what I should insert.

phlash:
> `size_t images = cv:imcount(path, cv:IMREAD_ANYCOLOR)`

I see a couple of issues with this approach:
 * `imcount` isn't available in OCV 4.5.1 (current on Debian stable), it appears in 4.5.3+
 * the implementation (https://github.com/opencv/opencv/blob/4.5.3/modules/imgcodecs/src/loadsave.cpp#L661) reads the whole input, assuming a file, which may hang for video streams or other URL backgrounds..

My existing implementation is a reduced form of this function, reading two frames to detect a video source. May I suggest we leave the existing mechanism alone, to get your work here merged?

jjsarton:
# Improvements:

## Detect kind of background file

The detection video or still image ist done with:

* the
number of images returned by cv::imcount() which is always
0 for video and gif file
* the mumber of frames for the possible video file

We have the following cases:

File type|Frame count|Image count
:--------|:---------:|:---------|
Video|> 1|0
Animated GIF|> 1|0
image|< 0|> 0
Static GIF|1|0
Not usable| < 1|0

## Opencv with ffmpeg

No further changes needed

## Opencv with gstreamer

Gstreamer don't respect the loop indicator, without the check if
we have read the last frame, the stream will be terminated.

Within the read thread the frame number is checked against the
number of frame for the "video". If we have read all frames the
stream position is set to 0. The result is not checked, if we read
past the last frame, the stream is closed and the corresponding
check for read error is always present.

## Not fixed behaviour

* Gstream may return a wrong frame rate for some GIF files.
* Gstream don't accept some files wich work well with ffmpeg.

This must not been fixed, the user shall use a correct file.

## Tests

Test were performed with various more files as such delivered with
backscsrub.
One version was tested with the opencv package provided for fedora 36:
```
Opencv version: 4.5.5
Video I/O:
    DC1394:                      YES (2.2.6)
    GStreamer:                   YES (1.20.0)
    v4l/v4l2:                    YES (linux/videodev2.h)
    Intel Media SDK:             YES (/usr/lib64/libmfx.so -lstdc++ /usr/lib64/libdl.a /lib64/libva.so /lib64/libva-drm.so)
```

The other version was an own compiled openvc:
```
Opencv version: 4.6.0-dev
Video I/O:
    DC1394:                      NO
    FFMPEG:                      YES
      avcodec:                   YES (59.18.100)
      avformat:                  YES (59.16.100)
      avutil:                    YES (57.17.100)
      swscale:                   YES (6.4.100)
      avresample:                NO
    GStreamer:                   YES (1.20.3)
    v4l/v4l2:                    YES (linux/videodev2.h)
```


Attached the result of git status against the last commit.
[background.cc.txt](https://github.com/floe/backscrub/files/9600538/background.cc.txt)

jjsarton:
I think that my code can remain how it is with the following exceptions:
instead of reading the number of images with imcount(), images is set to 0.
at the proper place I will set images to 1 if cnt < 0 (before video detection)
I have made a test on both backscrub version with all my backgrounds and had no error.

jjsarton:
We may conditionaly include cv::imcount().

If we have this after the #include lines, we have knowledge about the used opencv version.

```
#if CV_VERSION_MAJOR < 4 ||\
     CV_VERSION_MAJOR == 4 && CV_VERSION_MINOR < 5 ||\
     CV_VERSION_MAJOR == 4 && CV_VERSION_MINOR == 5 && CV_VERSION_REVISION < 3
# define HAVE_IMCOUNT 0
#else
# define HAVE_IMCOUNT 1
#endif
```
The modified code will be:
```
#if HAVE_IMCOUNT
        size_t images = cv::imcount(path.c_str());
#else
        size_t images = 0;
#endif
        pbkd->cap.open(path, cv::CAP_ANY);    // explicitly ask for auto-detection of backend
        if (!pbkd->cap.isOpened()) {
            if (pbkd->debug) fprintf(stderr, "background: cap cannot open: %s\n", path.c_str());
            return nullptr;
        }
        pbkd->cap.set(cv::CAP_PROP_CONVERT_RGB, true);
       ...
#if !HAVE_IMCOUNT
        if ( cnt < 0 )
             images = 1;
#endif
```
With this, the output messages are ok if the OCV version is high enough.

If OCV support only gstreamer and the version is too old, line

```
                if (pbkd->debug) fprintf(stderr, "background: imread cannot open: %s\n", path.c_str());
```
will be reached if a video file is not recognized. 

We can also modify the text to ``` "background: cannot open: %s\n"``` if we don't like the erroneous output, and don't use the conditional code.

phlash:
@jjsarton Thank you for the effort investigating the behaviour of OpenCV with various input files.. can I ask you to move that work (and related code changes) to a fresh PR against issue #158 please? It's not directly connected to your work here on croppping, and I'd like to get this work completed for the benefit of others, then we can spend more time on the format issue(s). Many thanks,

lib/libbackscrub.cc
phlash: This is now covered by issue #156, and thus can be fixed by a separate PR that properly deals with the fact that we do not account for models that have different input and output sizes at all.

app/background.cc
jjsarton: Size will not change, the frame content will change. Mofifying this implies:
if ! rect_calculated
  crop =  calc_cropping()
else
   crop = get_stored_crop_value()
...
I don't think that the calculation will take a lot ot time more than filling the cv::Rect from already calculated variables.

jjsarton:
A lot to do!
What shall be the base for this PR?
I will modify the file deepseg.c after the first PR is integrated within backscrub. There are some crashes while using the debug mode level 2. After this reworking, the argument parser will also be the subject of a new PR, but first if a discussion tell us what we have to do.
I will prefer that Bug fixes and detection optimization remain within cropping. Writing/modifying code mean also testing it. Test shall occur on different environment, so that any developer with another system may also detect possible bugs.

BenBE:
@jjsarton I personally prefer small and concise PRs with focussed changes. Currently this PR consists of >20 commits which I already asked previously to squash down into a more sensible set of commits. As this was not done yet I refrained from touching anywhere near the merge button for this PR. ;-)

If you have crashes or other bugs that are unrelated to the core issue of an PR, they should be separate. There's no problem with having multiple PRs in parallel aiming at specific aspects of the code. In fact: Doing so makes integrating them actually easier. If there are some dependencies for the order of PRs, just state them. Blowing up one PR over weeks doesn't make merging it any easier.

Marking this as draft for now, until the commit squashing has been resolved.

jjsarton:
@BenBE I Know, I had to learn, and my approach was not the best. The next PR I will prepare will only cover a problem and the corresponding code, so that the chance that there will not be a need to request many changes.

BenBE:
There's absolutely no problem if a PR accumulates many comments for changes that need to be done. In fact with any non-trivial change this is completely natural.

But once those are resolved would you rather read a commit history of 24 changes partially reverting each other or a concise set of about 5-10 commits "telling a story" of what needed to be changed to get to the solution?

Thus please have a look at [rewriting your commit history with `git rebase`](https://git-scm.com/book/en/v2/Git-Branching-Rebasing) and combine all those commits that fix typos or revert previous changes so the history of the commits in this PR is more manageable.

jjsarton:
@BenBe I has already wrote a reply to a comment like: [# 153](https://github.com/floe/backscrub/pull/153#issuecomment-1252413534) I will definitely do this, this IMHO not a good idea.

I have joined diff files against the floe/main and the (not pushed code) I would like to have.

The reason is that backscrub shall work on different distributions, not only Debian or Debian based.

Some Distribution have integrates ffmpeg within OCV, others don't do this.

The diff files contain all changes. deepseek.cc (6 changes),  libbackscrub.h (1 change) and libbackscrub.cc (2 changes). There include only the changes we have discussed.  
  
For background.cc, there are actually 6 changes. The diff is performed against a not pushed version and contain, if possible,
a good file type detection (cv::imcount() used if provided), one fix for GIF files while using gstreamer and finally the discussed suggestions.  

The file type detection as realized with the main sources is definitively worse. Reading a still image 2 times is possible with
gstreamer so that there are detected as video!.

The only subject of changes is normally (I assume that the other file are as wanted) background.cc.

[background.cc.txt](https://github.com/floe/backscrub/files/9617177/background.cc.txt)
[deepseg.cc.txt](https://github.com/floe/backscrub/files/9617178/deepseg.cc.txt)
[libbackscrub.cc.txt](https://github.com/floe/backscrub/files/9617179/libbackscrub.cc.txt)
[libbackscrub.h.txt](https://github.com/floe/backscrub/files/9617180/libbackscrub.h.txt)
[sources.tar.gz](https://github.com/floe/backscrub/files/9617181/sources.tar.gz)
[test.py.gz](https://github.com/floe/backscrub/files/9617182/test.py.gz)

